PR #14, opened 2025-01-24
Turn on feature accessibility input

Niduank:
* Display the "open app" shortcut on text selection
* Display the input view on text selection
* Handle app switching (multi text selection)
* Turned on Endpoint
* Display Input view in GeneratingContent & GeneratedContent (history)
* Resize the ContentView properly when switching apps

⚠️  We should test the local endpoint (code looks fine but I don't have ollama configured)
⚠️  I couldn’t find a way to highlights only code

### DIFF
--- a/macos/Onit/FeatureFlag/FeatureFlagManager.swift
+++ b/macos/Onit/FeatureFlag/FeatureFlagManager.swift
@@ -23,7 +23,7 @@ class FeatureFlagManager: ObservableObject {
     
     @Published private(set) var flags: FeatureFlags = .init()
     
-    struct FeatureFlags {
+    struct FeatureFlags: Codable {
         var accessibility: Bool = false
     }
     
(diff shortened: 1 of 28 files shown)


macos/Onit/FeatureFlag/FeatureFlagManager.swift
```diff
@@ -73,6 +73,10 @@ class FeatureFlagManager: ObservableObject {
         }
         
         flags = newFlags
+        
+        let preferences = Preferences.shared
+        preferences.featureFlags = newFlags
+        Preferences.save(preferences)
```
timlenardo, line 79: Feature flags are getting reset every time I restart the app. It looks like you're saving the overrides here. Can you make sure they're getting loaded each time?

timlenardo (changes requested):
Let's make sure the feature flags are persisted. Otherwise, lgtm!

Issue #21, opened 2025-01-26
Remote Ollama Endpoint


rapatel0:
I typically use ollama on a remote personal GPU server to avoid turning my laptop into a space heater. This works natively in ollama using the `OLLAMA_HOST` environmental variable. 

Can we have a config option to specify and ollama endpoint?

timlenardo:
Interesting! It's cold here, so I've been enjoying the laptop space heater. But I can understand why others might want something else. 

Is Ollama _running_ on your laptop? Or is Ollama running on your GPU server and there's a custom URL that you can hit from your laptop?

reestr:
I also have this requirement.  I have Ollama running on another system and it's accessed using a custom URL.

Lennaert:
Yes this one is something I also noticed. I have my Ollama on my GPU computer, and I can reach it in my local network. A Custom URL / host:port would be an ideal solution.

rapatel0:
I use ollama endpoint for a few apps but also the `ollama` cli tool respects the OLLAMA_HOST variable. 

I have a docker container running on my home server (with a beefy GPU) and then i have a vpn that puts in on the same network as my laptop (tailscale).

timlenardo:
PR for this is ready! @Niduank has a home setup with a remote PC and will confirm! 

@reestr @Lennaert @rapatel0 - if you are building a local version of the app, please validate as well 🙏

reestr:
@timlenardo - sorry I'm a bit slow following up..  I've built this locally and can confirm it is working as expected.  Thank you!

PR #27, opened 2025-01-28
Custom local endpoint

timlenardo:
Adding a customizable local endpoint URL per #21 

To achieve this, we needed to make some changes to Preferences - notably, we changed it from a Struct to a Class and only used a static MainActor instance of preferences. We also needed to update the local endpoints to pull a localEndpointURL from the preferences object. Finally, we add a field in Settings allowing users to set the localEndpointURL, and validate that it's a valid URL and that models can be loaded from it, with appropriate error messages in each failure case. 

@Niduank, can you test this on your home setup?

### DIFF
--- a/macos/Onit/App.swift
+++ b/macos/Onit/App.swift
@@ -43,6 +43,7 @@ struct App: SwiftUI.App {
         featureFlagsManager.configure()
         model.showPanel()
         
+
         #if !targetEnvironment(simulator)
         
         let hostingController = NSHostingController(rootView: StaticPromptView())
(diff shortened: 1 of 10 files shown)


sammcj:
I've tested the app built from this branch with my remote Ollama servers and can confirm it works as expected.

timlenardo:
Alright- given @sammcj's verification I'll merge this. @Niduank can you confirm as well?

Niduank:
I confirm it's working well with local ip and ngrok endpoint 🔥

PR #30, opened 2025-01-29
New user improvements

timlenardo:
I've added a few things here to  make the new user experience better. Notable, I've added a "setup Local models" and "setup remote models" button to the model dropdown, when the user does not have any local or remote models available. 

I've also changed the model dropdown to say "choose model" when there is no model set, instead of being blank. 

In general, the logic for keeping track of which model providers are validated and which model is selected is disparate and confusing, which means there are inevitably bugs. I'm seeing weird behavior in the edge cases (for example, when you submit an invalid token for a provider you had previously verified). We should do a bigger cleanup at some point in the future. 

Screenshot 2025-01-29 at 1 58 13 PM

Screenshot 2025-01-29 at 1 58 06 PM

### DIFF
--- a/macos/Onit/UI/Prompt/Selection/ModelSelectionView.swift
+++ b/macos/Onit/UI/Prompt/Selection/ModelSelectionView.swift
@@ -40,6 +40,7 @@ struct ModelSelectionView: View {
 
             if model.listedModels.isEmpty {
                 Button("Setup remote models") {
+                    model.settingsTab = .models
                     openSettings()
                 }
                 .buttonStyle(SetUpButtonStyle(showArrow: true))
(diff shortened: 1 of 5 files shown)


macos/Onit/UI/Prompt/Selection/ModelSelectionView.swift
```diff
@@ -38,7 +38,20 @@ struct ModelSelectionView: View {
             }
             .padding(.horizontal, 12)
 
-            remoteModels
+            if model.listedModels.isEmpty {
+                Button("Setup remote models") {
+                    model.settingsTab = .models
```
Niduank, line 43: Added auto-switch to Settings/Models tab

macos/Onit/UI/Prompt/Selection/ModelSelectionView.swift
```diff
@@ -78,7 +91,20 @@ struct ModelSelectionView: View {
             }
             .padding(.horizontal, 12)
 
-            localModels
+            if model.preferences.availableLocalModels.isEmpty {
+                Button("Setup local models") {
+                    model.settingsTab = .models
```
Niduank, line 96: Added auto-switch to Settings/Models tab

PR #35, opened 2025-01-31
Feature UI final context

Niduank:
### Auto Context

Added shortcut to launch with auto-context: CMD+Shift+0
Change shortcut to switch between local/remote becomes: CMD+Shift+L: @timlenardo feel free to change

### App sometimes freezes with Accessibility

Here’s what I’ve found so far :
* Removed 33 unused accessibility notifications, which means we avoid 33 observers each time an application is focused.
* The "code" rendering for the Input preview, which only works partially, uses the MarkdownUI library. When the selected text is too large, the CPU reaches 100% or even exceeds it if we repeatedly change the focus between apps. I turned off the code highlight for Input preview.
* The window which display the auto-context was blocking the main thread if the text was too long. I fixed it by adding a loader.
* The input preview was blocking the main thread if the text was too long. I fixed it by adding a loader.

### FeatureFlags

* I changed the way to use it, now we can override one or multiple feature flags.
* Now we have these feature flags working only if the "parent" feature flag is enable
  - accessibility Which turn on/off listeners
    - accessibility_input Which turn on/off the input feature (preview and hint)
      - highlight_hint_mode : textfield, topRight & other will be considered as "none" (no hint)
    - accessibility_autocontext which turn on/off the auto-context feature

To display the textfield hint you need : `accessibility == true && accessibility_input == true && highlight_hint_mode == textfield`

### Experiment

Everything is setup to launch the experiment but I can't create an Experiment on PostHog without the "control" variant, which is added by default and cannot have a 0% rollout.
@timlenardo I don't know what to do about this !

### Analytics
- `shortcut_launch` with properties `app_hidden`, `highlight_hint_visible` & `highlight_hint_mode`
- `shortcut_launch_with_auto_context` with property `app_hidden`
- `highlight_hint_mode_change` with properties `old` & `new`
- `app_quit`


timlenardo:
Hey @Niduank - awesome, looks good! For the Control variant, let's _not_ show the hint at all (i.e. neither .topRight nor .textField). That will be equivalent to the current production experience, so it will serve as a good control.

Niduank:
> awesome, looks good! For the Control variant, let's _not_ show the hint at all (i.e. neither .topRight nor .textField). That will be equivalent to the current production experience, so it will serve as a good control.

@timlenardo 
The current code will work as you want.
If the variant is different from 'textfield' or 'topRight', we will not display the hint.

What percentage do you want for the Control? 33% or less

timlenardo (approved):

PR #41, opened 2025-02-03
Replace UserDefaults with Defaults sdk

Niduank:
* All UserDefaults key are declared in `Defaults.swift`
* Removed `Preferences.swift` file
* Removed SageKit & files related to persistence
* Cleaning OnitModel
* Change all views using UserDefaults to support new SDK
* Added the "cleanup_app.sh" script to clean app data

### Reminder
- SwiftUI views should use the `@Default(.key)` annotation to support real time changes
- Outside SwiftUI view : 
  - If you need to read value in one shot use `Defaults[.key]`
  - If you need to get value in real time use `@ObservableDefault(.key)` & `@ObservationIgnored`


macos/Onit/UI/Prompt/Dialogs/SetUpDialogs.swift
```diff
@@ -21,6 +21,7 @@ struct SetUpDialogs: View {
     
     @AppStorage("closedRemote") var closedRemote = false
     @AppStorage("closedLocal") var closedLocal = false
+    @AppStorage("seenLocal") var seenLocal = false
 
     @AppStorage("closedOpenAI") var closedOpenAI = false
     @AppStorage("closedAnthropic") var closedAnthropic = false
```
timlenardo, line 27: @Niduank - I'm wondering if we should migrate the AppStorage variable over to our shared Defaults object too. What do you think?

timlenardo (approved):
Small question on AppStorage variables. Otherwise, looks good to me!

PR #46, opened 2025-02-05
WIP: Add custom providers

Niduank:
### Settings
We can add custom provider in `Settings` which will fetch models.
We can toggle a custom provider
We can toggle a model from custom provider
Data is persistent

### ModelSelection
We can see models from custom provider
We can select/use model from custom provider

### SetUp Remote Models (dialog and warning)
Models from custom provider are considered as remote models.
If we configure a custom provider, we consider that remote models are configured.

## Bugs
⚠️ Onit models can have same id than models from custom provider
- [x] Disable one in settings will disable all -> due to visibleModelIds
- [x] All model from custom provider are deprecated in Settings

### Testing
Let's set OpenAI & this custom provider in Settings
- name: "Test"
- url: "https://api.openai.com"
- token: "YOUR_TOKEN"

### Improvement
- [ ] Unselect from Settings the current model should pick a default one ?
- [ ] Find a way to skip model which doesn’t work on `chat/completion` ?


timlenardo:
I'm made a number of changes here, and as far as I can tell, this is working as expected. A summary of changes: 

- Changed CustomModelsResponse to work for response from other providers (we only require an 'id' now). I've tested this with OpenRouter and Groq. 
- Added uniqueId property to AIModel to handle model identification across providers
- Added a migration, so old versions of visibleModelIds are migrated to uniqueId instead of the original model ID
- Moved the CustomChatEndpoint logic into the main "chat" function with the other providers. 
- Updated the settings UI to include: a search bar, a scrollview, a TokenView (plus refresh), and a trash button to remove a custom provider. 

Some remaining tasks: 
- We aren't using the token validation logic, which I actually might prefer. 
- Oddly, through OpenHands, you can send SystemPrompts to models (like o1) that don't support system prompts. So, I'm including the system prompts in all the custom model chats, but this may not be the correct solution. We may need to let users enable/disable this on a per model basis.

PR #51, opened 2025-02-06
Streaming support

Niduank:
## Actual state
The streaming is enabled for local models and disable for remote ones.
You can enable it for remote models by modifying `OnitModel.useRemoteStreaming`

## Working with
- Local (ollama)
- OpenAI (not O1 https://community.openai.com/t/streaming-support-for-o1-o1-2024-12-17-resulting-in-400-unsupported-value/1085043)
- Anthropic
- GoogleAI
- xAI
- DeepSeek

## TODO:
- Check custom provider
- Improve `OnitModel.useRemoteStreaming` to switch between one-shot and stream


macos/Onit/Data/Model/OnitModel.swift
```diff
@@ -68,11 +68,15 @@ import SwiftUI
...
     var updater = SPUStandardUpdaterController(
         startingUpdater: true, updaterDelegate: nil, userDriverDelegate: nil)
+    var streamingClient = StreamingClient()
     var _tokenValidation = TokenValidationState()
 
     var remoteFetchFailed: Bool = false
     var localFetchFailed: Bool = false
-
+    
+    var useRemoteStreaming: Bool = false
```
timlenardo, line 77: Instead of having this be a global boolean, I'd like to make it a switch per provider. So there should be a "streamOpenAIResponses = true/false" and a "streamDeepSeekReponses = true", etc. There should also be an option for each Custom Provider. These should be editable in the Settings menu, so if you want to disable streaming (like if it doesn't work for a custom provider), you can do so there. It should default to true for the hardcoded providers (OpenAI, DeepSeek, Google, xAI, Anthropic), and default to false for any custom providers.

timlenardo: Oh also, we should have a toggle for streamLocalResponses also, for the Ollama endpoints, and default that to true!

macos/Onit/Data/Model/Model+Chat.swift
```diff
@@ -85,36 +86,77 @@ extension OnitModel {
...
+                                                                 apiToken: apiToken)
+                    }
+                case .local:
+                    guard let model = Defaults[.localModel] else {
+                        throw FetchingError.invalidRequest(message: "Model is required")
+                    }
+                    
+                    addPartialPrompt(prompt: prompt, instruction: curInstruction)
+                    
+                    let asyncText = try await streamingClient.localChat(systemMessage: systemMessage,
```
timlenardo, line 144: Add option for non-streamed response, if user enabled it in settings

macos/Onit/Data/Fetching/Endpoints/StreamingEndpoint.swift
```diff
@@ -0,0 +1,95 @@
+//
+//  StreamingEndpoint.swift
+//  Onit
+//
+//  Created by Benjamin Sage on 10/4/24.
```
timlenardo, line 5: 🧐

macos/Onit/Data/Fetching/Endpoints/Endpoint.swift
```diff
@@ -0,0 +1,91 @@
...
+                }
+                if let additionalHeaders = additionalHeaders {
+                    for (header, value) in additionalHeaders {
+                        print("  -H '\(header): \(value)' \\")
+                    }
+                }
+                print("  -d '\(jsonString)'")
+            }
+        }
+    }
```
timlenardo, line 90: This code is duplicated in Endpoint, StreamingEndpoint, and FetchingClient+Execute. Do we need it everywhere?

macos/Onit/Data/Fetching/Endpoints/ChatStreaming/OpenAIChatStreamingEndpoint.swift
```diff
@@ -0,0 +1,77 @@
...
+        }
+        
+        return nil
+    }
+    
+    func getStreamingErrorMessage(data: Data) -> String? {
+        let response = try? JSONDecoder().decode(OpenAIChatStreamingError.self, from: data)
+        
+        return response?.error.message
+    }
```
timlenardo, line 45: These functions are duplicated in all Streaming endpoints. Can we move them to the root "StreamingEndpoint" file?

timlenardo (changes requested):
Some small changes, but otherwise looks good!

macos/Onit/Data/Model/OnitModel.swift
Niduank: I totally agree with you but I didn't know how you wanted it.
Let's do what you said 👍

macos/Onit/Data/Fetching/Endpoints/StreamingEndpoint.swift
Niduank: Copy/Paste of `Endpoint.swift` 🤣

macos/Onit/Data/Fetching/Endpoints/ChatStreaming/OpenAIChatStreamingEndpoint.swift
Niduank: No we can't because we need it to fetch the text to display or the error depending on a specific Endpoint's response.

macos/Onit/Data/Fetching/Endpoints/Endpoint.swift
Niduank: I cleaned this part

macos/Onit/UI/Settings/Models/CustomProvider/CustomProviderRow.swift
```diff
@@ -12,13 +12,22 @@ struct CustomProviderRow: View {
...
 
     @State private var searchText: String = ""
     @State private var loading = false
     @State private var validated = false
     @State private var errorMessage: String?
     @State private var showAlert = false
+    
+    private var streamResponseBinding: Binding<Bool> {
+        Binding {
+            streamResponse.customProviders[provider.id] ?? false
```
Niduank, line 26: Default value for custom provider declared here

macos/Onit/UI/Settings/Models/ModelStreamResponse.swift
```diff
@@ -0,0 +1,25 @@
...
+//
+
+import SwiftUI
+
+struct ModelStreamResponse: View {
+    @Binding var isOn: Bool
+
+    var body: some View {
+        HStack {
+            Text("Use streaming response")
```
Niduank, line 15: Should be renamed

timlenardo:
I pulled this down and testing it. Looks good to me! 

I decided to make a few small changes:
- I moved the toggle to an 'advanced' section in settings, like we have for the local models. The "Models" settings page is getting quite crowded, so I think it's best to collapse this by default. This gives us a dedicated position where we can add more options like this in the future: 

Screenshot 2025-02-12 at 10 35 09 AM

- I added some explainer text so people can better understand the feature, and moved from a Switch to a RadioButton to better match the other settings: 
Screenshot 2025-02-12 at 10 35 21 AM

Finally, I renamed the component to the "StreamingToggle", which better describes what it is!

timlenardo (approved):
Let's get this merged! Thanks for all your hard work here @Niduank, it's looking great!

PR #53, opened 2025-02-06
Run swift format on entire codebase

jrswanson:
- Inconsistent whitespace is breaking the OpenHands Github action
- Swift comes packaged with a formatter
- Run `swift format . --recursive --in-place`


timlenardo (approved):
I scrutinized this heavily and, after much deliberation, determined that it looks good to me!

PR #70, opened 2025-02-12
System Prompts

Niduank:
### Old system message 
We re-use the old system message as a default non-editable and non-removable `SystemPrompt` called `outputOnly`.
`SystemPrompt.outputOnly` is inserted in SwiftData at app launch if there is no entry.

### Settings
List is sorted by timestamp desc
Search bar will filter on `SystemPrompt[.name, .prompt, .tags]` containing text
When deleting the selected SystemPrompt, we will try to select the most recently used one or `outputOnly`

### ContentView
We always display/use a system prompt

### Suggested System Prompt
The suggestions are displayed asynchronously.
Each SystemPrompt's suggestion is calculated based on a score.

For each application:
+5 for a match with the active (frontmost) application
+5 for a match with the input application (selected text)
+5 for a match with the applications from auto-context

For each tag:
+3 for a match in the input (selected text)
+3 for a match in auto-context and context files

## Misc
* New `KeyboardShortcutManager` which observe, enable/disable keyboard shortcuts
* Changes on `ModelContainer` & `EnvironmentValues`
* Script `clean_app.sh` updated - was not deleting default.store* files after backup


timlenardo:
Hi @Niduank - overall this looks great! I have one big behavioral change to request: 

Since the system prompts are a special case that are inserted only at the beginning of the chat messages, we shouldn't be able to change the system prompt once after the first message. The prompt should show up only once at the top of the chat, and not be editable in the "Follow-up" field:

Screenshot 2025-02-18 at 1 37 54 PM

To use a different system prompt, users will have to start a new chat. 

This might create a confusing experience, so I asked Elise to add a message explaining why the system prompt can't be edited (and a 'start new chat' button). Design is [here](https://www.figma.com/design/EP1rawvwdFlmO4DPSn7F4U/Onit-Product?node-id=2855-17815&m=dev)

Niduank:
> Hi @Niduank - overall this looks great! I have one big behavioral change to request:
> 
> Since the system prompts are a special case that are inserted only at the beginning of the chat messages, we shouldn't be able to change the system prompt once after the first message. The prompt should show up only once at the top of the chat, and not be editable in the "Follow-up" field:
> 
> Screenshot 2025-02-18 at 1 37 54 PM
> To use a different system prompt, users will have to start a new chat.
> 
> This might create a confusing experience, so I asked Elise to add a message explaining why the system prompt can't be edited (and a 'start new chat' button). Design is [here](https://www.figma.com/design/EP1rawvwdFlmO4DPSn7F4U/Onit-Product?node-id=2855-17815&m=dev)

I made the changes, everything works as expected. 
**Just to be sure, do we still need to send the system with each prompt?**
If so, feel free to merge it 👍

timlenardo:
Hi @Niduank - Yes, the system prompt gets sent with each prompt, but it's static and always at the top of the message stack. 

For example, if a user is sending a follow-up message, the stack will be:

System Prompt > 1st User Message > 1st LLM Reply > 2nd User message

timlenardo:
Hey Kevin - this looks great! There are a few changes to the header that should go along with this, as shown in [this design](https://www.figma.com/design/EP1rawvwdFlmO4DPSn7F4U/Onit-Product?node-id=2213-8382&m=dev).

Screenshot 2025-02-19 at 9 35 27 PM

Basically:
- The 'full screen' button moves to the top right
- The 'new chat' button moves to the left
- There should be an arrow next to the 'new chat' button
- By default, the system prompt UI should NOT appear. 
- If Onit has automatically _changed_ the system prompt (due to Application or Keyword or something else) it should show the system prompt UI. 
- If the user hovers over the down button next to the "new chat" button, the System prompt UI appears. 
- If the user clicks the down button next to the "New chat" button, there's a window to "start new chat with system prompt" and shows the interface for choosing a system prompt. 

Also, when the **first** time I ran this, I got a crash. It's worked on subsequent runs, but here's the crash I got: 

Screenshot 2025-02-19 at 9 32 28 PM

Niduank:
@timlenardo 
Changes made with your last feedbacks 👍

timlenardo:
@Niduank - two comments:

1st: I'm seeing a strange "bounce" animation after sending the first message. The loading indicator scrolls off-screen vertically and then bounces back when the result loads. I didn't notice this the last time I reviewed this PR, so I checked out the old code and traced the issue back to [32d3e3](https://github.com/synth-inc/onit/pull/70/commits/9fcd05090b2ac5f3100f8aa1e330965bff32d3e3). Can you fix this? 

https://github.com/user-attachments/assets/d409665a-d5e0-432e-b12e-50ee65f44adf

2nd: I occasionally see a crash "Trying to set a preview item on a closed preview view" on [line 39 of QLIMage](https://github.com/synth-inc/onit/blob/fa748a1e86b8549c66307092feb7f218c94473d7/macos/Onit/UI/Components/QLImage.swift#L39), implying that we are deleting the loading icon before the GIF loads. We can easily fix this by adding a check to make sure the view still exists before setting the previewItem. But I'm wondering, do you know what's causing this? I haven't seen this crash outside of this PR, so I'm wondering if it's caused by some of the updated rendering logic (like in the commit mentioned above).

timlenardo:
[Here's a PR](https://github.com/synth-inc/onit/pull/93) that should fix my 2nd comment. We can merge this if there isn't a clear cause for the issue from this PR!

timlenardo:
I made a small change to show the system prompt when the users clicks the "Start new chat" button in the system prompt explainer. I think this will lead to a better user experience, as the feature is currently pretty hidden!

https://github.com/user-attachments/assets/59d09368-d945-41d5-8e86-e69e4f2563b1

PR #80, opened 2025-02-17
Fix window resizing

Niduank:
* Fix a bug when ESCAPE is pressed with panel expanded
* Remove the padding on window
* Set maxHeight to ContentView
* Create a TextViewWrapper to limit the input's height correctly

### DIFF
--- a/macos/Onit/UI/Content/ContentView.swift
+++ b/macos/Onit/UI/Content/ContentView.swift
@@ -19,14 +19,6 @@ struct ContentView: View {
         )
     }
 
-    var maxHeight: CGFloat? {
-        if let height = NSScreen.main?.visibleFrame.height {
-            return height - 16 * 2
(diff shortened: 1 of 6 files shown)


macos/Onit/UI/Content/ContentView.swift
```diff
@@ -41,7 +33,7 @@ struct ContentView: View {
         }
         .background(Color.black)
         .buttonStyle(.plain)
-        .frame(minWidth: 325, idealWidth: 400)
+        .frame(minWidth: 325, idealWidth: 400, maxHeight: NSScreen.main?.visibleFrame.height ?? 0)
```
timlenardo, line 36: Let's leave some padding here for situations where the dock is not visible but could become visible!

timlenardo (changes requested):

PR #89, opened 2025-02-20
Fix issue #78: [Feature] Support editing of messages

### LINKED ISSUE #78: [Feature] Support editing of messages
It would be great to have the ability to edit every sent message. Re-entering a message will remove the older chat history and display a new response.

onitbot:
This pull request fixes #78.

The changes implement message editing functionality by:

1. Adding an `isEditing` flag to the Prompt model to track edit state
2. Modifying the send action logic to handle two scenarios:
   - For editing: Clears the existing responses and prior instructions, updates the instruction with new text, and regenerates
   - For new messages: Creates and generates a new prompt as before

The changes directly address the request to "edit every sent message" by:
- Providing a mechanism to identify which message is being edited
- Clearing old responses when a message is edited
- Regenerating the response with the updated text
- Preserving the original prompt object while updating its content

This implementation allows users to edit existing messages while maintaining the expected behavior of regenerating responses based on edited content, which matches the requirement that "re-entering a message will remove the older chat history and display a new response."

### DIFF
--- a/macos/Onit/Data/Persistence/Prompt.swift
+++ b/macos/Onit/Data/Persistence/Prompt.swift
@@ -23,6 +23,7 @@ import SwiftData
     var nextPrompt: Prompt?
 
     @Transient @Published var generationState: GenerationState? = GenerationState.done
+    @Transient @Published var isEditing: Bool = false
     var generationIndex = -1
 
     init(
@@ -35,6 +36,7 @@ import SwiftData
         self.contextList = contextList
         self.responses = responses
         self.generationState = GenerationState.done
+        self.isEditing = false
     }
 
     var generation: String? {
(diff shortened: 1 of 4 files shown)


timlenardo:
@onitbot - can you proceed and fix the remaining issues you listed above? 
❌ No implementation of regenerate button functionality to append instructions
❌ Missing logic for instruction editing triggering generate()
❌ Unclear how instruction history is being maintained

Regarding the 2nd point, there should be an edit button on the view that displays the instruction. When clicked, this should make the instruction editable. In the place of the edit button, there should be a "send" button and a "cancel" button. If they edit and click send, then we should trigger generate() again. This should append the new instruction the instruction array and increment the generationIndex.

timlenardo:
The bot didn't take the right direction on this one. Closing it out. #166 actually adds this feature!

PR #93, opened 2025-02-24
fix for loading indicator crash

timlenardo:
Small fix for a crash I've seen a few times now with the loading indicator. The error message is " Trying to set a preview item on a closed preview view," implying that we are setting the URL after the view has been deallocated. I add a nil check to prevent this.


Niduank:
Looks good 👍

PR #97, opened 2025-02-24
Fix issue #96: Swiping on Hint view should dismiss it.

### LINKED ISSUE #96: Swiping on Hint view should dismiss it.
If the user clicks and drags on the hint view, it should be dismissed. For the StaticPromptView, let's dismiss the view with a swipe to the right. For the OnitPromptView, let's dismiss the view on a swipe in any direction. 

To dismiss the StaticPromptView, we should use the opposite animation to display it. For OnitPromptView, we should dismiss in the direction of the drag.

onitbot:
This pull request fixes #96.

The changes fully implement the requested dismissal behaviors for both views:

For OnitPromptView:
- Added drag gesture support in any direction
- Dismisses when drag distance exceeds 50 points
- Animates dismissal in the direction of the drag using the drag angle
- Returns to original position if drag distance is insufficient
- Prevents tap-to-launch during drag

For StaticPromptView:
- Added right-swipe-only drag gesture support
- Dismisses when rightward drag exceeds 50 points
- Animates dismissal by sliding off screen to the right
- Returns to original position if drag distance is insufficient
- Prevents tap-to-launch during drag

The implementation matches all requirements:
1. Different dismissal behaviors for each view type
2. Direction-specific constraints (any direction for Onit, right-only for Static)
3. Appropriate animations (direction-based for Onit, right-slide for Static)
4. Proper threshold detection for gesture completion
5. Smooth animation handling with proper cleanup

The code is complete and should provide the exact user experience requested in the issue description.

### DIFF
--- a/macos/Onit/UI/OS/OnitPromptView.swift
+++ b/macos/Onit/UI/OS/OnitPromptView.swift
@@ -19,6 +19,11 @@ struct OnitPromptView: View {
 
     var body: some View {
         if model.panel == nil {
+            // Reset offset when view appears
+            let _ = DispatchQueue.main.async {
+                offset = .zero
+                isDragging = false
+            }
             HStack(spacing: 3) {
                 Image(.smirk)
                     .resizable()
--- a/macos/Onit/UI/OS/OnitPromptView.swift
+++ b/macos/Onit/UI/OS/OnitPromptView.swift
@@ -10,6 +10,8 @@ import SwiftUI
 
 struct OnitPromptView: View {
     @Environment(\.model) var model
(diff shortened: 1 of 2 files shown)


timlenardo:
@onitbot - in your implementation, when the StaticPromptView or OnitPromptView are dismissed, they never appear again. We need to make sure they can be displayed again. We should set Offset back to zero before displaying them again. I think this probably should in the adjustWindow() method, but am open to alternatives.

timlenardo:
We're moving towards a different hint view, so this is no longer relevant. Closing this out!

PR #101, opened 2025-02-27
wrap previews in debug conditional

timlenardo:
Small fix here: I can't archive the app for release unless these previews are wrapped in a DEBUG conditional. I'm not entirely sure why, but it's consistent with what we have elsewhere in the codebase. @Niduank do you know why?


Niduank:
@timlenardo This is because PreviewSampleData is inside the "Preview Content" folder which is builded only for development. And when you archive, Xcode is building the Previews (mistake on their side) ... So we should :
- wrapped the preview with DEBUG conditional when it use `PreviewSampleData`
- or put the sample data directly inside the Models

Issue #63, opened 2025-02-12
Context menu not closing


elisejansson:
In certain cases (unsure how to reproduce), the context menu freezes and cannot be dismissed unless quitting Onit completely.

timlenardo:
I experienced this issue this morning- very irritating. I'm not clear on the steps to reproduce, but attaching a screen recording here. 

https://github.com/user-attachments/assets/a52c9665-ab2f-4fad-bf2a-ef3c0e528b79

timlenardo:
This will be solved by [PR 102](https://github.com/synth-inc/onit/pull/102)!

PR #102, opened 2025-02-27
Route all overlays through a single controller

timlenardo:
This solves issue #63  

We have two different ContextOverlayControllers: one for the Model Selector and one for the Context Menu. There were race conditions when both were active: only one would consume the mouse-down events; and not always the right one.!

So, if you had the Context Menu active and tapped the select Model dropdown, sometimes the mouse-down event would get consumed by the activating Model Selector. So the Model Selector wouldn't get shown (or it would appear briefly and then immediately get killed by the mouse-down event). Because the Model Selector was consuming the mouse-down event, the Context Menu wouldn't get it. But the reference to the ContextOverlayController _would_ get deallocated. So we'd end up with a hanging view that was never dismissed, but there was no longer reference to it. 

The 'close' button on the hanging view wouldn't work,  because it routes the action through the model, which no longer had reference to the relevant ContextOverlayController. 

My solution is to route all ContextOverlayControllers through a shared OverlayProvider. The OverlayProvider will make sure that any prior ContextOverlayController is dismissed before launching a new one. This means only one can be listening for the MouseDown events at any given time, and it's always the right one. 

Moving forward, **we'll have to be conscious of listening for these mouse-down events. Any time we have more than one active listener, there will be race conditions**! 

@Niduank -tagging you so you're aware of the issue.

### DIFF
--- a/macos/Onit.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/macos/Onit.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -6,8 +6,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/google/abseil-cpp-binary.git",
       "state" : {
-        "revision" : "194a6706acbd25e4ef639bcaddea16e8758a3e27",
-        "version" : "1.2024011602.0"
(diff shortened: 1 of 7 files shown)


Niduank:
@timlenardo Noted. Indeed, these are constraints specific to macOS that don't exist on iOS. I'll be more careful.

PR #120, opened 2025-03-04
Fix issue #116: Entire Panel Position Button should be clickable

### LINKED ISSUE #116: Entire Panel Position Button should be clickable
In the "Panel Position" area in Settings, only the icon and words are clickable. The whole surrounding area should be clickable. Basically, everything that is purple here should be clickable:

Image

onitbot:
This pull request fixes #116.

The changes directly address the clickable area issue by restructuring how the background and interaction areas are defined in the SwiftUI view hierarchy:

1. Previously, the background color and clip shape were nested inside the content VStack, which meant only the inner content was effectively clickable
2. By moving `.background()` and `.clipShape()` modifiers to be applied directly to the entire button view, the full rectangular area (including the purple background) becomes part of the interactive button surface
3. Adding `.buttonStyle(.plain)` ensures the button behaves consistently
4. The explicit `.background(panelPosition == position ? Color(.blue300) : Color.clear)` ensures there is always a background layer to click on, even in the unselected state

These changes maintain the same visual appearance while expanding the clickable area to match the entire colored rectangle shown in the screenshot. The restructuring of the view modifiers means the hit testing will now work on the full button area rather than just the icon and text, which directly solves the reported issue.

### DIFF
--- a/macos/Onit/UI/Settings/GeneralTab.swift
+++ b/macos/Onit/UI/Settings/GeneralTab.swift
@@ -141,14 +141,10 @@ struct GeneralTab: View {
                                 }
                                 .frame(maxWidth: .infinity)
                                 .padding(.vertical, 8)
-                                .clipShape(RoundedRectangle(cornerRadius: 6))
-                                .background {
-                                    if panelPosition == position {
-                                        Color(.blue300)
-                                    }
-                                }
                             }
                             .buttonStyle(.plain)
+                            .background(panelPosition == position ? Color(.blue300) : Color.clear)
+                            .clipShape(RoundedRectangle(cornerRadius: 6))
                         }
                     }
                 }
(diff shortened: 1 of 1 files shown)


timlenardo:
OnitBot didn't get this one right. The correct solution was to add ".contentShape(Rectangle())" to the Button label. 

I'm a bit surprised that OnitBot failed here, since I got the correct solution from Sonnet 3.7 in Cursor. @jrswanson - let's update OnitBot to use 3.7 instead of 3.5.

timlenardo (approved):

timlenardo:
Demo of the updated functionality (for the multi-modal models that end up training on this 😅)

https://github.com/user-attachments/assets/f81bcd94-668b-4eca-a851-881dfdaef798

PR #152, opened 2025-03-21
Fix Multi-Monitor Awareness

lk340:
# Task

* Onit was soft-crashing on multi-display setups.
* We were previously using `NSScreen.main`, which is a single-screen property that can change unexpectedly during monitor switches.
* `NSScreen.main` has been replaced with `panel.screen` which returns the specific NSScreen where the window panel is currently located and `NSScreen.screens`, which is an array that accounts for all available displays, giving Onit better control over which screen should be taken into account.
* `NSScreen.screens.first` is being used as a fallback so that Onit will always at least capture the first connected display in the array. `NSScreen.screens.first` also accounts for single-monitor setups.


timlenardo (approved):
LGTM!

PR #156, opened 2025-03-21
Add cancel generation feature

timlenardo:
This PR adds a "Cancel Generation" button, for the case where you send a message and realize it was wrong, but don't want to wait until the generation finishes before fixing it. This is particularly helpful with the reasoning or web search models, which can take a long time to finalize a response. 

I won't merge this PR until we get some cleaner designs, but I wanted to get @Niduank's review on some of the changes I made. 

Screenshot 2025-03-21 at 3 56 11 PM

### DIFF
--- a/macos/Onit/Data/Model/Model+Chat.swift
+++ b/macos/Onit/Data/Model/Model+Chat.swift
@@ -244,9 +244,37 @@ extension OnitModel {
     func cancelGenerate() {
         generateTask?.cancel()
         generateTask = nil
+        
         if let curPrompt = generatingPrompt, let priorState = generatingPromptPriorState {
+            // Restore the prompt's previous state
             curPrompt.generationState = priorState
+            
+            // If this is the most recent prompt (last in the chat), remove it from the chat history
+            if let currentChat = currentChat, let lastPrompt = currentChat.prompts.last, lastPrompt.id == curPrompt.id {
+                // Get the instruction before removing the prompt
+                let instruction = curPrompt.instruction
+                
+                // Remove the prompt from the chat
+                currentChat.prompts.removeLast()
+                if let currentPrompts = currentPrompts, !currentPrompts.isEmpty {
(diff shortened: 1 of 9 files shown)


macos/Onit/Data/Model/Model+Chat.swift
```diff
@@ -341,7 +384,7 @@ extension OnitModel {
         prompt.generationIndex = (prompt.responses.count - 1)
         prompt.generationState = .done
         
-        generatingPrompt = nil
-        generatingPromptPriorState = nil
+//        generatingPrompt = nil
+//        generatingPromptPriorState = nil
```
timlenardo, line 388: @Niduank - I had to comment out these lines to get the cancellation to work. These functions, "addPartialPrompt" and "updatePrompt" get called when the streaming request starts. If we set generatingPrompt and generatingPromptPriorState to nil, we can't restore the prompt when a cancellation happens. 

Do we need these lines here? Why did you originally set it up so that we reset these whenever these functions are called?

Niduank: @timlenardo 
No we don't need it, I think I made a mistake while developing the streaming support.

Niduank:
@timlenardo 
I found a bug
Step to reproduce - When cancelling a streaming task which didn't received any response.
What's happened - The loader stay displayed.

timlenardo:
This one needs a lot of work still... 

- Instead of using generationState .done when responses are streaming, we should introduce some sort of intermediate state, like .streaming. 
- On cancellation when a request is still in the .streaming state, we should delete the Response object, and sometimes the Prompt and Chat objects, if it's the first of either. 
- We should duplicate this logic for request retries or Prompt Editing. This needs to get rebased on main, due to #166  which introduces Prompt Editing. 
- Pending some updated designs

timlenardo:
Outdated!

PR #162, opened 2025-03-25
Web Context

lk340:
# Task

* Add web context to Onit.
* URLs are detected in the main text input view and added as web context items.
* URLs are parsed based on whether they're PDFs or not, turned into text files, and then appended to the web context items for LLMs to have direct context into the web contents.
* Web context persists throughout conversation history.


timlenardo (commented):
Things to do: 
- You can get in a weird state if you send the message before the webpage is scraped. I think we should either block send action, or show a warning message if the user tries to send the prompt before the web scraping is done. 

- When you click on the context tag, we should show an overlay window that contains the content, like we do for AutoContext elements. 

- I think we should remove the URL from the TextInputView (and instruction) after it's parsed into a webContext tag.

macos/Onit/UI/Components/TextViewWrapper.swift
```diff
@@ -119,9 +124,90 @@ struct TextViewWrapper: NSViewRepresentable {
...
+                     for url in urls {
+                         let urlExists = parent.model.pendingContextList.contains { context in
+                             if case .web(let existingUrl, _) = context {
+                                 return existingUrl == url
+                             }
+                             return false
+                         }
+                         
+                         if !urlExists {
+                             parent.model.pendingContextList.append(.web(url, nil))
```
timlenardo, line 193: This should get routed through the Model+Input addContext() function, instead of hitting the list directly

macos/Onit/UI/Components/TextViewWrapper.swift
```diff
@@ -119,9 +124,90 @@ struct TextViewWrapper: NSViewRepresentable {
                 parent.cursorPosition = textView.selectedRange().location
             }
             
+            handleUrlDetection(text: textView.string)
+
```
timlenardo, line 128: We need to be able to disable this, since I'm now re-using this TextViewWrapper for editing prompts (in #166). In that context, it won't make sense to detect URLs and add them as context, since you can't change the context once you've already sent the message!

macos/Onit/UI/Components/TextViewWrapper.swift
```diff
@@ -119,9 +124,90 @@ struct TextViewWrapper: NSViewRepresentable {
                 parent.cursorPosition = textView.selectedRange().location
             }
             
+            handleUrlDetection(text: textView.string)
+            
             updateHeight()
         }
         
+        private func detectURLs(in text: String) -> [URL] {
```
timlenardo, line 132: This  *kinda* feel like this should be in a helper class somewhere, rather than directly in the TextWrapperView, but I think it's okay to leave it here for now and refactor later.

macos/Onit/UI/Components/TextViewWrapper.swift
timlenardo: Also, I'd like to delete the URL characters from this TextField when it's added as context!

macos/Onit/UI/Prompt/Files/WebContextItem.swift
```diff
@@ -0,0 +1,176 @@
...
+        }
+        .onAppear() {
+            fetchWebpageFavicon()
+            
+            if isEditing {
+                Task {
+                    await withTaskGroup(of: Void.self) { group in
+                        group.addTask { await fetchWebpageTitle() }
+                        group.addTask { await fetchWebpageContents() }
+                    }
```
timlenardo, line 92: So I think these tasks should get created and managed in Model+Input 'addContext",  rather than in this view. Right now, the chain of causality is: 

1. The user types a URL in TextViewWrapper, which detects it and adds a Context item to pendingContextList. 
2. The PromptView is listening to that list, and creates this View. 
3. This View triggers the fetching of the webpage contents and holds the "Tasks". 

This causes problems because the 'x' button to close the Context is in ContextItem. That calls model.removeContext, which is in Model+Input. Neither of those have access to these 'Tasks', so we can't cancel the tasks if the 'x' button is clicked to remove context before the scraping is done. 

Instead, it should be: 
1. The user types a URL in TextViewWrapper, which detects it and calls addContext in Model+Input.
2. addContext triggers these tasks, holding reference to them. AddContext also adds the new context element to pendingContextList. 
3. TextInputView is listening to the pendingContextList, so it creates this view. But this view is "dumb" and just listens for updates from Model+Panel tasks, rather than managing them itself. 
4. fetchingWebpageTitle, fetchingWebpageContents bools should either be on the contextItem or should be in the model. Probably in the model, with an array that associates webpage URLS with whether or not they're currently being fetched. 
5. I would put the title in the ContextItem. .web(URL: remoteURL, String: title, URL: localContentURL)
6. I think we can leave 'faviconFetching' here since it's so simple. 

This approach detangles the web content fetching from the WebContextItem view, which is good because I won't use this same UI to implement web search!

macos/Onit/UI/Prompt/Files/WebContextItem.swift
```diff
@@ -0,0 +1,176 @@
...
+        }
+    }
+    
+    func fetchWebpageFavicon() {
+        if let webpageDomain = url.host() {
+            faviconUrl = URL(string: "https://\(webpageDomain)/favicon.ico")
+        }
+    }
+    
+    func fetchWebpageTitle() async {
```
timlenardo, line 104: Don't we get the title when we get the full web content? I think we could probably just do one request here instead of two. 

If the goal is to make the title show up faster, we can always get it _first_, update the 'fetchingWebpageTitle' to true which will reset this UI, and then do the more expensive scraping afterwards. But that can all be done in a single function: no need for two. That SwiftSoup.parse call should be expensive, so we should avoid doing it twice!

macos/Onit/UI/Prompt/TextInputView.swift
```diff
@@ -22,6 +22,9 @@ struct TextInputView: View {
     
     @State private var textHeight: CGFloat = 20
     private let maxHeightLimit: CGFloat = 100
+    
+    @State private var detectedURLs: [URL] = []
+    @State private var urlDetectionTask: Task<Void, Never>? = nil
```
timlenardo, line 27: Dead code? If so, remove!

timlenardo (changes requested):

macos/Onit/UI/Components/TextViewWrapper.swift
timlenardo: This causes a weird bug where, when you're editing a prompt, you can paste a URL and it gets parsed as context, but only for new prompts, not the one you're editing. I'm going to disable this for the instance of TextViewWrapper that is for prompt editing, but leave it on in the main TextInputView. We can resolve the issue later, as part of sharing re-using the TextInputView for both sending initial messages and editing existing ones!

macos/Onit/UI/Prompt/Files/WebContextItem.swift
```diff
@@ -0,0 +1,101 @@
...
+    //
+    @State private var fetchingWebpageFavicon: Bool = false
+    @State private var fetchingWebpageTitle: Bool = false
+    @State private var fetchingWebpageContents: Bool = false
+    
+    var fetchingWebpageData: Bool {
+        return fetchingWebpageFavicon || fetchingWebpageTitle || fetchingWebpageContents
+    }
+    
+    init(item: Context, websiteUrl: URL, websiteTitle: String, isEditing: Bool) {
```
timlenardo: I don't think we should pass these in since they're already in the context item? seems like it could be annoying to re-use if we need a switch statement every time. Let's do the extraction of URL and title in this class instead!

macos/Onit/UI/Prompt/Files/WebContextItem.swift
```diff
@@ -0,0 +1,101 @@
...
+                    }
+                }
+                .appFont(.medium13)
+            }
+        }
+        .opacity(beingScraped ? 0.5 : 1)
+        .disabled(beingScraped)
+    }
+    
+    private func getCurrentWebsiteTitle() -> String {
```
timlenardo, line 78: This functions seem wonky to me. Why do we need to loop through all the of the contexts to get the title when it's already passed in as a parameter?

timlenardo (approved):

macos/Onit/UI/Prompt/Files/WebContextItem.swift
timlenardo: Makde this change in b1202e

macos/Onit/UI/Prompt/Files/WebContextItem.swift
```diff
@@ -0,0 +1,101 @@
...
+    private let websiteTitle: String
+    private let isEditing: Bool
+    //
+    @State private var fetchingWebpageFavicon: Bool = false
+    @State private var fetchingWebpageTitle: Bool = false
+    @State private var fetchingWebpageContents: Bool = false
+    
+    var fetchingWebpageData: Bool {
+        return fetchingWebpageFavicon || fetchingWebpageTitle || fetchingWebpageContents
+    }
```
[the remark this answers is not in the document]
timlenardo: I made this change in b1202e

macos/Onit/UI/Prompt/Files/WebContextItem.swift
timlenardo: @lk340 - we don't have to solve this now, but curious to get your take on why we need this function?

PR #181, opened 2025-04-04
Undo Chat History Deletion

lk340:
# Task

* [Designs](https://www.figma.com/design/EP1rawvwdFlmO4DPSn7F4U/Onit-Product?node-id=4296-12002&t=25TV19GrpIfRqRX4-0)

* Add ability to undo history deletion.

# Notes

* The current window for undoing chat history deletion is 3 seconds before the chance to undo expires.
* Added slight modifications to original design for potentially better UX. Looking for feedback on this.
  1. Can undo multiple deletions in the case of rapid consecutive deletions.
  2. Add chat names to differentiate different deletion cards on consecutive deletions.
  3. Add decaying progress bar (a new component that can be customized in various ways) for a visual indication on undo countdown.
* Applied fixes to some text truncation issues in history view rows.


timlenardo (changes requested):
Requesting changes due to the issues I've mentioned in Slack: 
- If you’re in an active chat, and delete that chat from history, the chat still stays open in the OnitPanel. I found that weird. I think it should probably disappear and start a new chat.
- If I delete a chat, it’s still accessible in history via the up and down arrows.
- I saw one weird case where I deleted a chat from history, then relaunched the app. The chat I deleted was still in my history, but a different chat had lost its _responses_. I haven't been able to reproduce this, but I do think there's something funky going on with the deletion that needs to be debugged. 
 
I've looked at the code, and the delete() function seems correct, so I suspect the issues are coming from the timeout, and storing the Chat to be deleted in chatQueuedForDeletion. Is it possible this becomes the wrong Chat? chatQueuedForDeletion is only a single Chat, so what happens if you delete a bunch simultaneously? Or, is it possible that the HistoryRowView gets recycled and the wrong  Chat gets selected? Either way, can you test this extensively to see if you can trigger these edge cases? Some of my testing is coming back fishy 🤨

PR #182, opened 2025-04-07
Add Incognito Mode

lk340:
### Task

- Add an “incognito mode” feature that allows users to have auto-context-and-history-exempt conversations.

### Notes

- Add a `incognito_mode_enabled` event for PostHog.


macos/Onit/UI/Prompt/Toolbar.swift
```diff
@@ -242,6 +244,19 @@ struct Toolbar: View {
...
+        Button {
+            incognitoModeEnabled.toggle()
+            model.newChat()
+        } label: {
+            Image(incognitoModeEnabled ? .incogOn : .incogOff)
+                .renderingMode(.template)
+                .padding(4)
+                .foregroundColor(incognitoModeEnabled ? .blue400 : .gray200)
+        }
+        .tooltip(prompt: "Incognito mode (auto-context off)")
```
timlenardo, line 258: This shouldn't say "auto-context off" since incognito just means it doesn't get added to history

timlenardo (approved):
LGTM- one small copy update but otherwise great!

PR #200, opened 2025-05-08
Force Onit to use Dark appearance

Niduank:


timlenardo (approved):
Works for me. This prevents very bad outcomes (i.e. black text on black bg), but we still want to be sure to always set custom colors.

PR #201, opened 2025-05-08
Auth and billing

lk340:
Auth and billing with Stripe and browser login integrations.


macos/Onit/UI/Alerts/TwoWeekProTrialEndedAlert.swift
```diff
@@ -0,0 +1,59 @@
...
+    
+    private func downgradeToFreePlan() async {
+        do {
+            descriptionActionLoading = true
+            
+            let client = FetchingClient()
+            try await client.updateSubscriptionCancel(cancelAtPeriodEnd: true)
+            await refreshSubscriptionState()
+            
+            descriptionActionLoading = false
```
jrswanson, line 53: This is a bit different than how the state logic works on the server. If their two week trial is over (they proceeded from the `trialing` state directly to the `canceled` state), setting `cancelAtPeriodEnd: true` won't do anything or might even error out. The downgrade has already happened essentially.

macos/Onit/UI/Settings/GeneralTabPlanAndBilling.swift
```diff
@@ -0,0 +1,314 @@
...
+                        }
+                    }
+                } else if appState.subscriptionStatus == SubscriptionStatus.trialing || appState.subscriptionStatus == SubscriptionStatus.active {
+                    manageSubscriptionButton
+                } else if appState.subscriptionStatus == SubscriptionStatus.canceled {
+                    HStack(spacing: 11) {
+                        renewSubscriptionButton
+                        manageSubscriptionButton
+                    }
+                }
```
jrswanson, line 86: Once a subscription has entered the cancelled state, it can't be renewed. They will have to create a new subscription.

jrswanson (changes requested):

timlenardo:
I'm going to go ahead and merge this to create our release-candidate build!

PR #209, opened 2025-05-14
Bug Fix: Auth Flow

lk340:
This PR addresses the following updates:

* Onit now automatically defaults to ***remote mode*** when logging in.
* Remove the following, as they are no longer needed due to users having access to remote models via the free tier: 
  * "Set Up Remote Models" dialog card
  * "Setup remote models" button
  * Warning icon from remote models selection header
  * Remote models selection view is now always present.
* Onboarding Auth:
  * Email validation is now executed on the submission level, rather than in real time.
  * Create custom Apple login button for handling Apple authentication to maintain [design parity](https://www.figma.com/design/EP1rawvwdFlmO4DPSn7F4U/Onit-Product?node-id=6480-20523&t=CHdT8wWeetKcx8tX-0) (out-of-box Apple sign in button cannot be styled to this degree)
  * Center text in "Continue with Email" button.
  * Add underline hover states to text buttons.
  * Reintroduce bordered Onit logo.
  * Fix spacing.


macos/Onit/UI/Content/ContentView.swift
```diff
@@ -118,9 +119,14 @@ struct ContentView: View {
             handleFileImport(result)
         }
         .addAnimation(dependency: state.showChatView)
+        .onChange(of: appState.account) {
+            if appState.account != nil {
+                mode = .remote
+            }
+        }
```
timlenardo, line 126: This doesn't feel like the right place for this. If this is the logic we want, I'd rather do this when we're handling the loginResponse. But I'm actually not sure this is the right logic. I think what we really want is to _not_ default to local mode for new users if they don't have API tokens setup, which I think can be handled elsewhere.

timlenardo (approved):
Made it through the onboarding successfully 🧘‍♂️ Let's get this merged into main so we can keep stress testing!

PR #210, opened 2025-05-14
Main window selection waterfall

timlenardo:
This is to replace findFirstTargetWindow(), which works a surprisingly well, but fails in some cases. My selection waterfall is as follows:

1. A single main window from target windows
.isMain returns true for one window for all the apps I've tested. This is optional though, so we have other fallbacks:

2. The application's mainWindow if it's in our target windows
.mainWindow() returns the correct window for the apps I've tested, but is also an optional attribute. 

3. The application's focusedWindow if it's in our target windows
.focusedWindow() also returns the correct window for apps I've tested. but is also an optional attribute. 

4. The first window in the target windows list
I'm not sure about this. In situations where the 'main' or 'focused' window is not a target window (like if you've selected the Google Hangouts overlay), it may be better to return Nil and not create the target window. What do you think @Niduank ?

### DIFF
--- a/macos/Onit/Accessibility/AXUIElement+Accessors.swift
+++ b/macos/Onit/Accessibility/AXUIElement+Accessors.swift
@@ -287,6 +287,39 @@ extension AXUIElement {
     func bringToFront() {
         AXUIElementPerformAction(self, kAXRaiseAction as CFString)
     }
+
+    public func frontmost() -> AXUIElement? {
+        if let value = self.attribute(forAttribute: kAXFrontmostAttribute as CFString) {
+            return (value as! AXUIElement)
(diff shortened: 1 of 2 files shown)


Niduank:
@timlenardo Awesome, I tested it and it works well.

I think for the #4 we should return nil instead of a random window (first one) which can causes bugs (resize/move/bring-to-front the wrong window).

I would like to implement it in my [accessibility PR](https://github.com/synth-inc/onit/pull/211) because I see that in cases #2 and #3, you're introducing the bug where accessibility won't be fully activated. You're accessing AXUIElementCreateApplication without retrieving the role attribute.

PR #212, opened 2025-05-15
Remote Model Bug Fixes

lk340:
With the new plan-based remote model limitations, the old method of preventing access to remote models via API-key restriction is now rendered defunct. With that in mind, this update brings the following: 

1. All remote models fetched from our server are now available and accessible within Settings → Models, regardless of whether the user has provided valid API keys for them or not.
2. All remote models fetched from our server are now available for selection and use within the model selection view, unless their respective providers have been turned off or if individual models have been unselected in settings.
3. All remote provider toggle buttons in settings are now always visible, as remote models are no longer gated behind the previously-required API token.

This update also brings a fix to an edge case where users are upgrading to the new version of the Onit. This fix turns all provider toggle defaults in Settings → Models to "on" for users who are upgrading.

### DIFF
--- a/macos/Onit/AppState.swift
+++ b/macos/Onit/AppState.swift
@@ -230,17 +230,13 @@ class AppState: NSObject {
                 TokenManager.token = loginResponse.token
                 account = loginResponse.account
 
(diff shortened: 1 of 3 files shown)


timlenardo:
Looks good to me!

PR #218, opened 2025-05-16
Bug Fix: Token Validation

lk340:
This PR fixes a critical bug where API tokens were not properly checked when sending or updating prompts.


timlenardo (approved):
Makes sense, let's get this merged!

PR #219, opened 2025-05-16
Update: Arrow Hotkeys Breaking User Experience

lk340:
This PR addresses the user-breaking-experience issue where the up, down, right, and left arrow hotkeys constantly interfere with prompt creation.

The arrow hotkeys are now disabled whenever the user is in the middle of typing or moving their cursor, debounced to 1000ms (subject to change).

### DIFF
--- a/macos/Onit/UI/Panels/State/OnitPanelState.swift
+++ b/macos/Onit/UI/Panels/State/OnitPanelState.swift
@@ -192,7 +192,7 @@ class OnitPanelState: NSObject {
         isTypingDebounceTask?.cancel()
         
         isTypingDebounceTask = Task {
-            try? await Task.sleep(for: .milliseconds(1000))
+            try? await Task.sleep(for: .milliseconds(2500))
             
             if !Task.isCancelled {
                 await MainActor.run { self.isTyping = false }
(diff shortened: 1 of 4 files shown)


timlenardo (approved):
I'm okay with this implementation, although I wonder if:
- There's a better way than a 1sec debounce timer
- 1 sec is the right amount of time. Should it be longer? 
- Will this just confuse users, because sometimes up/down/left/right does one thing and other times it does something else, without a clear indication in the interface. 

Let's get this merged for now, but we should talk about it as a team at the next sync!

timlenardo:
I'm going to make the debounce timer a bit longer (1.0 sec -> 2.5 sec) and merge this. That's an arbitrary taste-based call—I pause sometimes when I'm typing and think it'd be nice to have it last longer than 1 second. I'm not sure, though. We'll have to play with it and discuss it as a team, and then we can make a call.

timlenardo:
@Niduank - what do you think of this one? I had the idea that we could observe the firstResponder status in the TextView itself, but that's _always_ true when the panels open so it doesn't work. Any other ideas?

Niduank:
@timlenardo @lk340 
### Context / Problem
When typing in a TextView, arrow key shortcuts no longer work. To avoid relying on debounce mechanisms, we want to inform the user only when:
* The TextView is focused
* The TextView is not empty

### Proposed UX Solution
Display an informational message and a button below the input, like this:
> "Arrow key shortcuts are disabled while typing. [Click here](https://github.com/synth-inc/onit/pull/219#) to exit input mode."

### Behavior of the "Click here" button
When tapped, it will programmatically remove focus from the TextView, re-enabling keyboard shortcuts.

PR #220, opened 2025-05-16
Remove Onit server toggle

jrswanson:
- In my first pass on the auth business logic, I imagined using Onit server chat completion would be user toggleable.
- On further reflection, we decided that we should just default to using any valid token the user inputs and fallback on the Onit server in its absence.
- This state logic does not represent the logged out state. From now on, users must create an account to use the interface.


timlenardo:
This seems reasonable to me. I am going to merge it and #216  into auth-bug-fixes so that I can test it all together!

PR #213, opened 2025-05-15
Feature/on demand ax parsing

Niduank:

PR #221, opened 2025-05-16
Parse AX tree on demand

timlenardo:
This is just a replay of @Niduank's PR #213 , which got caught up in a weird merge state. I've rebased it on the most recent main, confirmed that it works, and am just creating a clean PR so the merge goes successfully. 

* Stop listening notifications related to the automatic parsing
* Adding a window's context is handled by StateManagers
* Remove the automaticallyAddAutoContext feature


Niduank:
@timlenardo About the old branch `feature/on-demand-ax-parsing`. 
I'm removing my local and the remote branches as it's already merged.

timlenardo:
@Niduank - yup, that sounds right!

PR #222, opened 2025-05-16
disable the pinned vs tethered toggle for users who havent given acce…

timlenardo:
…ssibility access

This is a small fix. I realized that we should be disabling the toggle between Pinned mode and Tethered mode if the user hasn't given Accessibility permissions: 

Screenshot 2025-05-16 at 3 53 13 PM

Screenshot 2025-05-16 at 3 53 23 PM

### DIFF
--- a/macos/Onit/UI/Settings/GeneralTab.swift
+++ b/macos/Onit/UI/Settings/GeneralTab.swift
@@ -2,6 +2,7 @@ import Defaults
 import PostHog
 import ServiceManagement
 import SwiftUI
+import AppKit
 
 struct GeneralTab: View {
     @Environment(\.windowState) private var state
@@ -17,6 +18,8 @@ struct GeneralTab: View {
     
     @State var isLaunchAtStartupEnabled: Bool = SMAppService.mainApp.status == .enabled
     @State var isAnalyticsEnabled: Bool = PostHogSDK.shared.isOptOut() == false
+    
+    @ObservedObject private var accessibilityPermissionManager = AccessibilityPermissionManager.shared
     
     var isPinnedMode: Bool {
         usePinnedMode ?? true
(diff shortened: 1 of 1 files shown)


macos/Onit/UI/Settings/GeneralTab.swift
```diff
@@ -65,11 +68,14 @@ struct GeneralTab: View {
     }
 
     var displayModeSection: some View {
-        Section {
+        let accessibilityGranted = accessibilityPermissionManager.accessibilityPermissionStatus == .granted
```
Niduank, line 71: @timlenardo 
Why did you not use a computed property for readability and avoid the `return Section` ?
```
private var accessibilityGranted: Bool {
    accessibilityPermissionManager.accessibilityPermissionStatus == .granted
}
```

macos/Onit/UI/Settings/GeneralTab.swift
```diff
@@ -65,11 +68,14 @@ struct GeneralTab: View {
...
+        return Section {
             VStack(alignment: .leading, spacing: 16) {
                 HStack(spacing: 8) {
                     Button {
-                        FeatureFlagManager.shared.togglePinnedMode(true)
+                        if accessibilityGranted {
```
Niduank, line 76: Not necessary as you're using the `disabled` view modifier, it's redundant

macos/Onit/UI/Settings/GeneralTab.swift
```diff
@@ -85,11 +91,15 @@ struct GeneralTab: View {
...
                         .clipShape(RoundedRectangle(cornerRadius: 6))
                         .contentShape(Rectangle())
+                        .opacity(accessibilityGranted ? 1.0 : 0.5)
                     }
                     .buttonStyle(.plain)
+                    .disabled(!accessibilityGranted)
                     
                     Button {
-                        FeatureFlagManager.shared.togglePinnedMode(false)
+                        if accessibilityGranted {
```
Niduank, line 100: Not necessary as you're using the `disabled` view modifier, it's redundant

Niduank (commented):
Working well 👍

Niduank:
@timlenardo I did all the changes so it can be merged

PR #226, opened 2025-05-19
Filter chats by account

Niduank:
* Save accountId in Chat
* First filter Chats by accountId then by StateManager


timlenardo:
Hey @Niduank - looks good, but two quick items: 

1. The current chat stays after a logout/login. So if I write a prompt from my synth.inc account, log out, and then log in with my gmail.com account, the prompt that I just sent from the other account is still visible in the panel:
Screenshot 2025-05-20 at 10 08 23 AM

2. Will this require a migration for existing users? I seem to have lost my history when I pulled this branch.

timlenardo (changes requested):
See previous comment!

macos/Onit/UI/Settings/AccountTab.swift
```diff
@@ -364,5 +364,10 @@ struct AccountTab: View {
     func handleLogout() {
         TokenManager.token = nil
         appState.account = nil
+
+        // Reset all chat state
+        for windowState in PanelStateCoordinator.shared.states {
+            windowState.newChat(clearContext: true)
+        }
```
timlenardo, line 371: Note: We aren't currently using AccountTab, so this update wasn't strictly necessary. But, I decided to do it anyway, in case we re-introduce the AccountTab. We should delete this file if we aren't going to!

macos/Onit/UI/Settings/GeneralTabAccount.swift
```diff
@@ -131,6 +131,11 @@ extension GeneralTabAccount {
         TokenManager.token = nil
         appState.account = nil
         Defaults[.authFlowStatus] = .showSignIn
+        
+        // Reset all chat state
+        for windowState in PanelStateCoordinator.shared.states {
+            windowState.newChat(clearContext: true)
+        }
```
timlenardo, line 138: @Niduank - this is my suggestion for how to resolve the issue we discussed in team meeting this morning. Let me know if you have a better idea!

timlenardo:
@Niduank, I left some comments on my edit here. Let me know what you think!
I will merge this to include it in my build this afternoon. If you want to make any changes, we can open a new PR with updates!

macos/Onit/UI/Settings/AccountTab.swift
Niduank: Totally agree. 
The old UI views was also a pain when working on the Coordinator, so we should definitely remove them.

PR #229, opened 2025-05-20
Prioritize Highlights when multiple context exists

timlenardo:
This is a simple fix, inspired by a bug report from Elise. When multiple types of context exists - i.e. both a HighlightedText and an AutoContext - we should be giving priority to the Highlighted text. This PR accomplishes that with: 

- Move the Highlights to the end, so they're the last thing before the user prompt
- Call out explicitly that highlights should be given priority. 
- Add more explanations about the files/other context. 

I've tested it in a few situations myself and it seems to work!

Screenshot 2025-05-20 at 12 52 08 PM

Note* This was a big problem in the prior iteration, where autocontext was automatically pulled constantly. It will be less of an issue now that users must add the autocontext explicitly. But, it's still good to fix!

### DIFF
--- a/macos/Onit/UI/Prompt/GeneratedErrorView.swift
+++ b/macos/Onit/UI/Prompt/GeneratedErrorView.swift
@@ -17,6 +17,7 @@ struct GeneratedErrorView: View {
                 Text(errorDescription)
                     .appFont(.medium14)
                     .foregroundStyle(.warning)
+                    .textSelection(.enabled)
             }
             Spacer()
         }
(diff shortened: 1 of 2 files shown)


macos/Onit/UI/Prompt/GeneratedErrorView.swift
```diff
@@ -17,6 +17,7 @@ struct GeneratedErrorView: View {
                 Text(errorDescription)
                     .appFont(.medium14)
                     .foregroundStyle(.warning)
+                    .textSelection(.enabled)
```
timlenardo, line 20: Minor thing- but it's helpful to be able to copy this so I can copy any error messages and search for them in the codebase. I added this as debugging code- it's unrelated to the PR/commit. However, I think it's good to get in the codebase for future use, so I left it here.

PR #231, opened 2025-05-20
Audit and reduce setup dialogs

timlenardo:
It seems like the team all agrees that these setup dialogs are mostly annoying, and not that useful:

Screenshot 2025-05-20 at 1 39 10 PM

From watching a handful of users onboard, I've been surprised by how many people entirely ignore these ! Now that setup isn't required (i.e. they don't have to input a token), we should remove as many of them as possible. 

This PR: removes some and also makes it so that the local-mode related setup dialogs only show up if/when you try to put the app into local mode. Demo here: 

https://github.com/user-attachments/assets/86361904-e5b9-404a-a612-d69416bd6b74


timlenardo:
I'm leaving in the 'model deprecated' and 'new model' dialogs for now, since it's essential info, and we don't have another way to communicate it. However, let's remove those too, once we find a different, less intrusive design to accomplish that same task!

PR #232, opened 2025-05-20
Show panel when dock icon is clicked

timlenardo:
This change makes the panel open (if it isn't already) when the user clicks on the dock icon or double clicks on the app in Applications. It has different behavior in the different modes: 

- In Tethered mode, it will open the panel on the current active application window, if it's not already open. 
- In Pinned mode, it will open the panel on the mouse screen, if it's not already open. 

I think this sort of thing is important as we support more users that are less technical, and are less likely to use shortcuts. When people click on an app icon in the Dock, they expect that app to open!

@Niduank - let me know what you think and/or if there are better ways to accomplish this with PanelStateCoordinator.


Niduank:
@timlenardo I made a few changes:

- Removed the Task which was unnecessary (the AppDelegate runs on the main thread) and, as greptile mentioned, could cause race conditions if the user spams clicks.
- Modified the coordinator to ensure no panel is open (to be discussed with you ⚠️ ).
- Ensured that currentState is never the defaultState and never launch the panel on the defaultState ⬇️ 

This fixes a bug when switching from Pinned to Tethered mode and click on Onit in Dock. It was using the defaultState, which meant a panel not attached to any app and that couldn't be closed.

Niduank:
@timlenardo If you're wondering why we have a defaultState even though we can't use it, to avoid having an optional currentState and having to add conditions everywhere.

timlenardo:
@Niduank on your second point _"Modified the coordinator to ensure no panel is open (to be discussed with you ⚠️ )",_ :

That's not how I was envisioning it working. Basically, in Tethered mode, it's easy to lose track of panels. So you could have a panel open, but it's attached to some app that's minimized in the background. Your new code would prevent a new panel from opening in this case, which could create a confusing user experience. I think the prior implementation is more similar to what a user would expect: clicking the Dock icon should open a new panel on the foregrounded application!

Niduank:
@timlenardo Got it! I put back the old condition that checks whether the current panel is not open. However, I kept the condition that ensures the current panel doesn't match the defaultState, because without it, there's a bug where the panel opens and never closes.

This happens when switching from pinned mode to tethered mode (no hint is visible, so no panel can be opened → defaultState).

PR #235, opened 2025-05-21
add logging around GoogleSignIn failures

timlenardo:
I spent all day trying to debug the 'keychain error' issue with GoogleSignIn. It seems to be working now, but I don't have a lot of confidence, since I couldn't understand the root issue. So, I'd like to add these analytics events so that we can track if/when users are running into this. If we see it in production, we will have to continue debugging or remove Google Sign-In until their team finds a consistent solution.


Niduank:
@timlenardo I already did it in the analytics PR.

timlenardo:
Sounds good, I will close this PR then!

PR #236, opened 2025-05-22
Fix Broken Auth Funnel & Accessible ChatView When Alerts Are Open

lk340:
* Fix for onboarding accessibility constantly reappearing every time the panel is reopened.
* Refactor onboarding in `ContentView`, since auth flow now exists outside of onboarding.
* Fix for `ChatView` being interactive when alerts are open.


macos/Onit/UI/Content/ContentView.swift
```diff
@@ -14,12 +14,25 @@ struct ContentView: View {
...
+        let accessibilityNotGranted = AccessibilityPermissionManager.shared.accessibilityPermissionStatus != .granted
+        return accessibilityNotGranted && Defaults[.showOnboardingAccessibility]
+    }
+    
+    private var showingAlert: Bool {
+        showTwoWeekProTrialEndedAlert || appState.showFreeLimitAlert || appState.showProLimitAlert
+    }
+    
+    private var showToolbar: Bool {
+        !ContentView.shouldShowOnboardingAccessibility && appState.account != nil
```
Niduank, line 33: I'm not sure ! As `shouldShowOnboardingAccessibility` is static we can lose the real time update on :
- accessibilityPermissionStatus (we don't care as the panel will be close by Coordinator)
- Defaults[.showOnboardingAccessibility] I think this one might be a bit annoying for you.

macos/Onit/UI/Content/ContentView.swift
```diff
@@ -43,9 +71,13 @@ struct ContentView: View {
...
                             
-                            if state.showChatView { ChatView().transition(.opacity) }
-                            else { Spacer() }
+                            if state.showChatView {
+                                ChatView().transition(.opacity)
+                            } else {
+                                Spacer()
+                            }
                         }
+                        .allowsHitTesting(!showingAlert)
```
Niduank, line 80: This is not fixing the bug where the alert is displayed with textfield focused.
To avoid the textfield to be focused when showing an alert we have to do:
```
.onChange(of: showingAlert) { _, newValue in
    state.textFocusTrigger = !newValue
}
```

https://github.com/user-attachments/assets/41d0553c-d4c8-4e7f-869b-cad403552cfb

Niduank (changes requested):

PR #239, opened 2025-05-24
Refacto - Remove unused code

Niduank:
Removed all the unused code except for the parts that might be useful to us in the future (commented out).
I used [Periphery](https://github.com/peripheryapp/periphery?tab=readme-ov-file#xcode-integration) tool to detect unused code.


timlenardo:
@Niduank, great idea 👍. Let's get this merged so we can ensure we haven't introduced any bugs before the next release!

PR #240, opened 2025-05-24
Implement secure token storage with AES-GCM encryption

Niduank:
- Add encrypted token storage in Application Support directory
- Implement AES-GCM encryption/decryption with 256-bit key
- Add proper error handling with custom TokenError type (not visible in UI)
- Use atomic writes with complete file protection

### DIFF
--- a/macos/Onit/Data/Fetching/TokenManager.swift
+++ b/macos/Onit/Data/Fetching/TokenManager.swift
@@ -42,9 +42,37 @@ struct TokenManager {
         }
     }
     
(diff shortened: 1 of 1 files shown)


timlenardo:
@Niduank  - I like the idea, but this PR will **log out all existing users**, since there will be no token at the location. I'd prefer to set it up so that existing users can keep using the token in Keychain, but new users will use the AES-GCM token. Can you make that change?

timlenardo:
@Niduank if you don't have objections, I'm going to merge this so we can include it in the beta release!

PR #243, opened 2025-05-29
Open panel on fresh start

timlenardo:
## Summary
- launch the panel automatically once the app has finished launching

## Testing
- `swift --version`
- `swift build -c release` *(fails: Could not find Package.swift)*
- `xcodebuild -list -workspace macos/Onit.xcodeproj` *(fails: command not found)*


timlenardo:
THis one is from OpenAI's codex and doesn't appear to work, so you can ignore it.

PR #246, opened 2025-05-30
Fix web search shimmer position

timlenardo:
## Summary
- prevent the searching indicator from jumping around
- add a new `FixedShimmer` modifier and use it instead of `.shimmering()`

## Testing
- `swift --version`


timlenardo:
So I asked OpenAI's codex to work on this problem and this is what it came up with. The approach doesn't look right to me, so let's ignore it for now.

PR #247, opened 2025-05-30
Fix IME phonetic input bug.

lk340:
Fix for adding in IME text inputs. Texted with Japanese Romaji and Chinese Pinyin.

Notes: 

- From testing, I’ve noticed that the audio recorder functionality is broken in that the lack of a valid API key causes issues with how the audio recorder adds/removes spacing in the text input view to make room for the `AudioRecorder`. Therefore, I’ve removed (commented out) the audio recorder button (`MicrophoneButton` in `PromptCoreFooter`) for now until we resolve the issue on the server.
- I’ve also done a look-see on any other external features that rely on the code block that is now gated within the `audioRecorder.isRecording || audioRecorder.isTranscribing || !textView.hasMarkedText()` conditional in `updateNSView`, and I believe that the only feature that was relying on that logic was the audio transcription feature, so this update shouldn't break anything. Please double-check @timlenardo

### DIFF
--- a/macos/Onit/UI/Components/TextViewWrapper.swift
+++ b/macos/Onit/UI/Components/TextViewWrapper.swift
@@ -80,19 +80,21 @@ struct TextViewWrapper: NSViewRepresentable {
         textView.audioRecorder = audioRecorder
         
         if textView.string != text {
-            let selectedRanges = textView.selectedRanges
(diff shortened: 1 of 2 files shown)


timlenardo (approved):
Awesome, thanks!

PR #248, opened 2025-05-30
Integrate server voice transcription

lk340:
Integrate Jay's server update for voice transcription


timlenardo:
I'm making a small change here. When I was testing, I got an index out of bounds error in the "addSpacesAtCursor()" function. I wasn't able to reproduce it - it's probably some uncommon race condition - so I decided to add a check and make sure that we aren't going out of the bounds. Otherwise, this looks good to me! Let's get this merged!

PR #253, opened 2025-06-02
Unfocus text input when app goes into background

lk340:
It's confusing to see a focused state on the text view (gradient border) when Onit is in the background. For example, if I'm currently on Chrome, Onit's text view should look inactive (no gradient border), but when clicking back into Onit, the gradient border should re-appear.

To resolve this, I've added an app-wide notification event to listen for when the Onit panel is active/inactive.

@Niduank One concern I have with my implementation is your comment in `OnitPanelState`:

> TODO: KNA - Unused function which can create bugs with keyboard shortcut

I'm wondering if this were referring to eventually removing the use of `panelResignKey` in `OnitPanelStateDelegate`.


timlenardo (changes requested):
Hey @lk340 - I think there's a better approach here, since we already have a **panelBecomeKey** and **panelResignKey** methods in the OnitPanelStateDelegate protocol.

Instead of using global notifications through NotificationCenter we should: 

1. Extend OnitPanelStateDelegate in PromptCore
2. Add it as a delegate to OnitPanelState in PromptCore's init (i.e. windowState.addDelegate(self)) 
3. Set isFocused=true in panelBecomeKey and isFocused=false in panelResignKey.

PR #259, opened 2025-06-03
New Text Highlight Experience

lk340:
* [Design File](https://www.figma.com/design/EP1rawvwdFlmO4DPSn7F4U/Onit-Product?node-id=8139-5217&p=f&t=sFmDG6svkISGI4To-0)

---

This PR brings in new updates to the text highlighting experience.

The old experience is still used by default, but users can opt-in through the "Context" tab in settings, titled "Highlighted Text: Multi-Context" (tentative titling), which also includes info details.

By opting in, highlighted text can be added as individual pieces of context, as opposed to having only one highlighted text input in the opted-out version of the text highlighting experience. Clicking on text context items reveals its contents in a scrollable Input view.

I've also added small UI updates to better match consistency with new designs.


macos/Onit/Data/Fetching/Streaming/StreamingClient.swift
```diff
@@ -54,14 +57,17 @@ actor StreamingClient {
...
                    responses: [String],
                    model: String) async throws -> AsyncThrowingStream<String, Error> {
         let userMessages = ChatEndpointMessagesBuilder.user(
             instructions: instructions,
             inputs: inputs,
             files: files,
             autoContexts: autoContexts,
-            webSearchContexts: webSearchContexts)
+            webSearchContexts: webSearchContexts,
+            textsContexts: textsContexts
```
timlenardo, line 69: Adding 'textContents' here is a duplicate of "inputs". We shouldn't have to abstractions for the same idea. Instead, since we want to support multiple inputs now (with Pinning), we should make the inputs parameter an array of arrays: `[[Input?]]` 

That way it will match the rest abnd we won't have to make all these changes in the data layer.

timlenardo (changes requested):
Hey @lk340 
Okay, after thinking this a bit, I’m okay with shipping (and not including a toggle to revert to the prior experience) but I'd like to change the implementation a bit. 

This PR should be split into three PRs: 

1. **PR 1** should add a ContextTag for highlightedText, in addition to the InputView. There should be a single toggle called “showInputView” to determine whether the InputView is shown. This value is set by user actions: if the user closes the InputView (by clicking the ‘x’), set it to false and only show the ContextTag for future highlights. If the user opens the InputView (by clicking on the ContextTag), set it back to true. It should also be possible to toggle this setting in the settings menu. In this PR, Inputs should always be added, just as they are now. No changes should be made to the internal logic (such as OnitPanelState, FetchingClient, StreamingClient, etc).
2. **PR 2** should add an optional default setting called ‘autoAddHighlightedText’ to control whether Inputs are added automatically. When this option is disabled, we should use the ‘ghost tag’ approach, as we do with context. This PR should introduce a single new field to OnitPanelState to track the current Input that has not yet been added. 
3. **PR 3** should introduce pinning by changing ‘pendingInput’ from a single Input to an array of Inputs. This PR will update the main internals (such as FetchingClient, StreamingClient, etc), but only by changing the existing “Input” parameter to accept an array, rather than adding a new “textContents” field.

Implementing it this way avoids adding confusing branching logic to the data layer and prevents new fields from being added to OnitPanelState. From a user experience, I’ll be able to use something similar to the current experience by setting .showInputView and .autoAddHighlightedText to true. Elise will be able to get the opposite effect by setting .showInputView and .autoAddHighlightedText to false.

PR #264, opened 2025-06-05
Fix - Audio Recorder UI getting clipped

Niduank:
@timlenardo 
The real issue came from :
1.  `CustomTextView.intrinsicContentSize`, which didn’t take the `RecordingIndicator` view into account
2. RecordingIndicator height was 22px, we were using 20px for the host's frame.

The ideal height seems to be 18px. We can increase it (the view won’t be clipped) but it will increase the size of the prompt (see video with maxHeight = 22).

https://github.com/user-attachments/assets/885bf2db-10ee-47cd-803d-167c364466c5


timlenardo (approved):
LGTM!

PR #266, opened 2025-06-09
Add automatic autocontext loading in tethered mode

timlenardo:
## Summary
- add a `autoContextOnLaunchTethered` default
- include new toggle in settings when using tethered mode
- watch the new flag for analytics
- automatically fetch window context when opening the panel in tethered mode

## Testing
- `swift build` *(fails: Could not find Package.swift)*

------
https://chatgpt.com/codex/tasks/task_b_68432d0d9240832fa0ea00e1f7d93393

### DIFF
--- a/macos/Onit/App.swift
+++ b/macos/Onit/App.swift
@@ -27,6 +27,7 @@ struct App: SwiftUI.App {
     @Default(.launchOnStartupRequested) var launchOnStartupRequested
     @Default(.autoContextFromCurrentWindow) var autoContextFromCurrentWindow
     @Default(.autoContextFromHighlights) var autoContextFromHighlights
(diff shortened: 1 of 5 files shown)


timlenardo:
This is another one requested by users. Two users have expressed confusion that context is not automatically added when an Onit Panel is opened in tethered mode. I decided to add it as a toggle in settings, since some other users may still prefer to not add context by default. 

I've confirmed this is working as expected! When enabled, autocontext will automatically add when you opened a tethered window. 
Screenshot 2025-06-08 at 9 42 43 PM

Here is the language in settings (it only appears in Tethered mode): 
Screenshot 2025-06-08 at 9 42 27 PM

PR #270, opened 2025-06-10
Context Menu

lk340:
Design Files:
* [New Context Menu](https://www.figma.com/design/EP1rawvwdFlmO4DPSn7F4U/Onit-Product?node-id=9117-8314&p=f&t=2EocR8TeWHUQ9r2V-0)
* [Actions From Input Field](https://www.figma.com/design/EP1rawvwdFlmO4DPSn7F4U/Onit-Product?node-id=3289-51317&p=f&t=ABq7xLuyaE2TwtWp-0)

---

This PR introduces the new Context Menu, which builds on window context by capturing all active windows and files, allowing users to easily traverse through captured entities to add and remove context items in quick succession.

The context menu also serves as a convenient window into seeing which windows are/aren't in context.

Keyboard traversal (up/down arrows and enter key), on-open active search bar, and auto-hovering of first option allow for immediate ease of use.


macos/Onit/Accessibility/Notifications/AccessibilityNotificationsManager.swift
```diff
@@ -269,18 +281,31 @@ class AccessibilityNotificationsManager: ObservableObject {
     
     // MARK: Parsing
     
-    private func retrieveWindowContent(for pid: pid_t, state: OnitPanelState?) {
+    private func retrieveWindowContent(
+        for pid: pid_t,
+        state: OnitPanelState?,
+        customAppBundleUrl: URL? = nil
+    ) {
         guard let mainWindow = pid.firstMainWindow,
```
timlenardo, line 289: @lk340  - We can't use "firstMainWindow" here. We need to find the specific window that correlates to the window we're trying to add. This is a bit tricky. We have a 'hash' concept in the TrackedWindow that partially works, but has some issues. We'll need to introduce a new function for finding a matching window and use it in a handful of places. We should discuss with @Niduank  in the eng sync tomorrow.

macos/Onit/UI/Panels/State/OnitPanelState+Input.swift
```diff
@@ -77,8 +85,93 @@ extension OnitPanelState {
...
+    
+    func addWindowToContext(
+        windowName: String,
+        pid: pid_t,
+        appBundleUrl: URL?
+    ) {
+        addAutoContextTasks[windowName]?.cancel()
+        
+        addAutoContextTasks[windowName] = Task {
+            guard let focusedWindow = pid.firstMainWindow
```
timlenardo, line 105: Same comment here: we need to pick the right window, not just the firstMainWindow. This logic introduces a lot of bugs in applications with many windows, like Chrome. For example, AutoContext items often get stuck in the 'loading' state because the added context doesn't have the same 'WindowName' as the AutoContext tag that was added. Try opening 5-10 Chrome tabs and then adding a few of them at once, and you'll see what I'm talking about. It stems from always selecting the "firstMainWindow" instead of choosing the right one.

macos/Onit/Accessibility/Notifications/AccessibilityNotificationsManager.swift
```diff
@@ -99,11 +99,23 @@ class AccessibilityNotificationsManager: ObservableObject {
...
+                state: state,
+                customAppBundleUrl: customAppBundleUrl
+            )
         } else if let pid = lastActiveWindowPid {
-            retrieveWindowContent(for: pid, state: state)
+            retrieveWindowContent(
+                for: pid,
+                state: state,
+                customAppBundleUrl: customAppBundleUrl
+            )
```
timlenardo, line 118: This function doesn't really do anything. This could just be a single line check at the beginning of 'retrieveWindowContext", something like: 
`guard let pid=pid ?? lastActivePID  `

macos/Onit/Accessibility/Notifications/AccessibilityNotificationsManager.swift
```diff
@@ -269,18 +281,31 @@ class AccessibilityNotificationsManager: ObservableObject {
     
     // MARK: Parsing
     
-    private func retrieveWindowContent(for pid: pid_t, state: OnitPanelState?) {
+    private func retrieveWindowContent(
```
timlenardo, line 284: @Niduank @lk340  - We should talk about a refactor of this code. It shouldn't all be in AccessibilityNotificationsManager, because Accessibility notifications no longer trigger it. We should introduce a "WindowContextManager" or something similar to handle parsing windows. 

I'd also like to separate the parsing logic into a protocol and have different implementations for the different strategies we use. (i.e. an AccessibilityWindowParser, an OCRWindowParser, a FileSystemWindowParser, etc etc.) Let's discuss in the team sync tomorrow.

macos/Onit/Accessibility/Notifications/AccessibilityNotificationsManager.swift
```diff
@@ -422,10 +456,15 @@ class AccessibilityNotificationsManager: ObservableObject {
...
+    private func handleWindowContent(
+        _ results: [String: String]?,
+        for state: OnitPanelState,
+        customAppBundleUrl: URL? = nil
+    ) {
+        var appBundleUrl: URL? = customAppBundleUrl
         
-        if let pid = lastActiveWindowPid,
+        if appBundleUrl == nil,
+           let pid = lastActiveWindowPid,
```
timlenardo, line 467: As an aside (below), it's weird to store the results in the 'screenResult' variable on the AccessibilityNotificationsManager. We should just pass the error and/or contents to the 'addAutoContext' function.

macos/Onit/Accessibility/Notifications/AccessibilityNotificationsManager.swift
```diff
@@ -363,7 +388,12 @@ class AccessibilityNotificationsManager: ObservableObject {
         return nil
     }
 
-    private func parseAccessibility(for pid: pid_t, in window: AXUIElement, state: OnitPanelState) {
+    private func parseAccessibility(
+        for pid: pid_t,
+        in window: AXUIElement,
+        state: OnitPanelState,
+        customAppBundleUrl: URL? = nil
+    ) {
```
timlenardo, line 396: We store the initial Task in OnitPanelState+Input, but then in the code below (lines 412 and 415), we dispatch it first onto a DispatchWorkItem and then into another detached Task. Cancelling that Task won't cancel the parsing, since it's taking place on a different, unrelated thread. The Task will look "completed" even though it's still processing. This also means our debouncing logic (which cancels the task if another request comes through for the same Window) won't work unless the new request comes before this dispatch happens. 

I'm unsure of the solution here, as we need this heavy processing to occur on a background thread. If we remove the DispatchWorkdItem, we can make the secondary Task and attached task, and that might work.

macos/Onit/UI/Prompt/Files/ContextMenuWindows.swift
```diff
@@ -0,0 +1,366 @@
...
+                )
+                
+                let notIgnoredApp = !checkWindowShouldBeIgnored(windowName)
+                
+                // Only capturing valid windows:
+                //   1. Has a PID.
+                //   2. Isn't the current window (this is handled by `currentForegroundWindowButton`).
+                //   3. Not in the list of ignored apps.
+                if let windowPid = window.pid(),
+                   windowPid != windowState.currentWindowPid,
```
timlenardo, line 348: This introduces a bug when you have multiple windows from the same app (such as five open Chrome windows). When a different app is foregrounded and you open the Context Menu, all of the Chrome windows appear as expected. But when Chrome is foregrounded and you open the Context Menu, only one Chrome window appears (first, via the currentForegroundWindowButton). This line of code prevents the other Chrome windows from appearing. 

Instead of matching on the pID, we need to match on the window specifically. We'll need to pass something to identify the current window into this function. We'll also need to choose how to best identify the current window, which introduces the same problem as above, where we don't have a reliable method for window

timlenardo (changes requested):

Niduank:
@timlenardo 
I completely agree on all three points:

* Since the new system for adding window contexts no longer relies on accessibility notifications, the parsing should be moved out of AccessibilityNotificationsManager.

* Regarding parsing, there’s already a strategy system in place based on the active application within `AccessibilityParser`. However, this system can be significantly improved to include detection of kAXDocumentAttribute or other parsing methods.

* Finally, we really need a reliable way to identify the correct window. The big issue with tabs (Chrome, Cursor, Arc browser) is that the windowHash is the same for all tabs! I haven’t found any accessibility property, aside from the title (which is dynamic), that can reliably distinguish them.

macos/Onit/UI/Prompt/Files/ContextMenuWindows.swift
```diff
@@ -0,0 +1,346 @@
...
+                capturedWindows = await captureOpenWindows()
+                
+                await MainActor.run {
+                    capturingWindowsTask = nil
+                }
+            }
+        }
+        .onDisappear {
+            capturingWindowsTask?.cancel()
+            capturingWindowsTask = nil
```
Niduank, line 190: Each time you're setting the `capturingWindowsTask` either to a new `Task` or `nil` you `cancel` it before.
You can add directly to the `capturingWindowsTask` property : 
```swift
@State private var capturingWindowsTask: Task<Void, Never>? = nil {
        willSet {
            capturingWindowsTask?.cancel()
        }
    }
```

macos/Onit/UI/Prompt/Files/ContextMenuWindows.swift
```diff
@@ -0,0 +1,346 @@
...
+
+// MARK: - Main Component
+
+struct ContextMenuWindows: View {
+    @Environment(\.windowState) private var windowState
+    
+    @Binding private var searchQuery: String
+    @Binding private var currentArrowKeyIndex: Int
+    @Binding private var maxArrowKeyIndex: Int
+    private let closeContextMenu: () -> Void
```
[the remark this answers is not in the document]
Niduank: Tim shared me this yesterday : https://philz.blog/mysterious-runtime-crashes-in-swift-6-language-mode/

macos/Onit/UI/Panels/State/OnitPanelState+Input.swift
```diff
@@ -77,8 +89,62 @@ extension OnitPanelState {
...
+                trackedWindow: trackedWindow,
+                customAppBundleUrl: appBundleUrl
+            )
+        }
     }
     
+    func getForegroundWindow() -> TrackedWindow? {
+        let windowsManager = AccessibilityNotificationsManager.shared.windowsManager
+        
+        if let trackedWindow = windowsManager.activeTrackedWindow {
```
Niduank, line 137: I saw a call to `AccessibilityNotificationsManager.shared.windowsManager.addToTrackedWindows()` in `ContextMenuWindows` which doesn't set the `activeTrackedWindow`. I think it can cause an issue

macos/Onit/Accessibility/Notifications/AccessibilityNotificationsManager.swift
```diff
@@ -382,28 +400,16 @@ class AccessibilityNotificationsManager: ObservableObject {
         let workItem = DispatchWorkItem { [weak self] in
             guard let self = self else { return }
             
-            Task.detached(priority: .background) { [weak self] in
-                guard let self = self else { return }
-                
+            Task {
```
Niduank, line 403: This task will not be cancelled if the DispatchWorkItem is cancelled.
But let it to me, I will do it in another PR.

Niduank:
**Test in tethered mode**

1. When minimizing an app:

    * It should no longer be possible to add it as a context window.
    * I've noticed that the panel attached to it sometimes doesn't disappear.

2. I'm not sure if this is a bug or the expected behavior, but when adding the context window of an app where text is selected, it correctly added the context window, but also included the selected text (this also happens on main).

3. When clicking on a panel, it brings the application it’s attached to to the foreground. However, the current context window doesn’t match (this also happens on main).

**Other suggestion**
Search in the window context menu should also support application names (users may not necessarily know the window title).

macos/Onit/Accessibility/Notifications/AccessibilityNotificationsManager.swift
timlenardo: Let's not change this because though, I will handle in a future PR.

PR #273, opened 2025-06-12
Save Tether hint position

Niduank:
* Work with all managers: Tethered, Pinned, Untethered.
* Untethered's tutorial follows the hint when dragged.
* Use relative positioning so panel's button follows the hint even when resizing app.


timlenardo:
Hey @Niduank - code looks good, but I'm not sure about the product experience here. I think we may want to save it on a per-application basis. For example, if you move it to the bottom every time you open Figma, it should remember that. 

I'm not sure what to do with Chrome/Arc/browsers. That's a unique case because there are many different windows at any given time. What do you think? I'm discussing with Elise and one of our users on Discord, and will update you with our thoughts.

timlenardo:
Let's ship this as is. I think it's okay to include it in Pinned mode for now. Additionally, it should probably persist across app launches, but we can address this in a future pull request.

PR #279, opened 2025-06-17
Ensure OnitPanelState instanciation on MainActor thread

Niduank:
@timlenardo Not sure about this one, if you can try to reproduce the bug, that would reassure me.


timlenardo:
Since we don't have a way to test this and aren't sure if it will resolve the issue, I won't merge this for now. I'm going to close it so it doesn't get accidentally merged, but we can revisit later if we can find more reliable repro steps for our crash.

PR #284, opened 2025-06-23
Fix duplicated windows in context menu.

lk340:
Add a fix for opening new windows resulting in the context menu being flooded with duplicates.


macos/Onit/UI/Panels/State/OnitPanelState.swift
```diff
@@ -51,7 +51,10 @@ class OnitPanelState: NSObject {
     var foregroundWindow: TrackedWindow? = nil {
         didSet {
             if let newWindow = foregroundWindow {
+                foregroundedWindowHistory.removeAll { $0 == newWindow.hash }
+                
                 foregroundedWindowHistory.append(newWindow.hash)
```
timlenardo, line 56: It doesn't really matter, but I feel like this is typically accomplished with an if statement that checks if the array contains it, and then only appends if it's not already included.

PR #285, opened 2025-06-23
Optimize recursive parsing of window context items

lk340:
@Niduank For context, this PR addresses something @timlenardo and I were in talks with about window context parsing getting heavy to process when trying to load in several at once. Namely, the UI would lag, freeze, or tasks would endlessly hang. This is my stab at trying to make improvements on this front.

---

### Task

Originally, when parsing heavy windows with tons of text - like multiple Wikipedia pages on countries or popular figures, for example - the UI would snag due to the load of recursively processing child nodes in one big operation on the main thread, per task.

This PR seeks to relieve such heaviness by chunking and yielding window node parsing with other operations that are shared on the same thread in short bursts, allowing the UI to function more fluidly.

---

### Testing

1. Open 8 Wikipedia pages with heavy text, in separate browser window instances (shift+left click):
    - https://en.wikipedia.org/wiki/Jordanian_annexation_of_the_West_Bank
    - https://en.wikipedia.org/wiki/Jordan_Peterson
    - https://en.wikipedia.org/wiki/Joran_van_der_Sloot
    - https://en.wikipedia.org/wiki/Jordan
    - https://en.wikipedia.org/wiki/Jordan_River
    - https://en.wikipedia.org/wiki/Jordan_Spieth
    - https://en.wikipedia.org/wiki/Joe_Biden
    - https://en.wikipedia.org/wiki/Joe_Rogan

2. Make sure that all Wikipedia pages are on the same desktop and that they all stretch to the max width/height of the screen.

3. Open the context menu. All Wikipedia pages should have loaded in as button options.

4. Click on all of the Wikipedia pages, one after the other. Their respective tasks should load in without freezing the UI, and consecutively loading parsed window contents into context should no longer result in split-second UI freezes for each added context item.

5. Repeat the above steps on `main` and compare with this branch. In my test for `main`, only a few windows would load into context while other window tasks would run endlessly until the app froze and gave me the dreaded beachball of doom 💀

---

### Side Note

I've moved `windowContextTasks[windowHash]` code from `addWindowToContext()` in `OnitPanelState+Input.swift` to `retrieveWindowContent()` in `AccessibilityNotificationsManager.swift` because it was unnecessarily wrapping a task inside of a task.

### DIFF
--- a/macos/Onit/Accessibility/Parser/AccessibilityParserUtility.swift
+++ b/macos/Onit/Accessibility/Parser/AccessibilityParserUtility.swift
@@ -45,8 +45,6 @@ class AccessibilityParserUtility {
         element: AXUIElement,
         maxDepth: Int,
         childNodeProcessingInterval: Int = 128,
-        // handler: (AXUIElement, Int) -> AccessibilityParseResult
-//        handler: @escaping (AXUIElement) async -> [String: String]?
         handler: @escaping (AXUIElement, Int) async -> AccessibilityParseResult
     ) async -> [String: String] {
         var results: [String: String] = [:]
@@ -80,10 +78,6 @@ class AccessibilityParserUtility {
              if let result = parseResult.results {
                  results.merge(result) { _, new in new }
              }
-            
-//            if let result = await handler(currentElement, currentDepth) {
-//                results.merge(result) { _, new in new }
-//            }
 
             // Only recursively parse children if handler didn't return .stopRecursing
             if parseResult.shouldContinue, let children = currentElement.children() {


macos/Onit/Accessibility/Parser/AccessibilityParser.swift
```diff
@@ -30,10 +30,11 @@ class AccessibilityParser {
         let parser = parsers[appName] ?? genericParser
         
         return try await withThrowingTaskGroup(of: [String: String].self) { group in
-            group.addTask { @MainActor in
+            // Using `@Sendable` here is okay, because we're allocating all operations to a single thread (main).
+            group.addTask { @MainActor @Sendable in
```
timlenardo, line 34: @lk340 did you add this because of the blog post that says you sometimes need @sendable even when it's a mainactor task?

PR #287, opened 2025-06-24
Fix toolbar overflow bug.

lk340:
### Task

Fix toolbar icons disappearing on small widths.

Main Files (these are the only files that explicitly need review):
* `ContentView.swift`
* `Toolbar.swift`

### Side Notes

In relation to this task, I took the liberty to make some long-awaited updates to the `IconButton` component:
* Previously, the clickable surface area was isolated to only the icon, making it annoying to click on sometimes. Now the entire button invokes the on-click callback.
* Tooltips are handled without repetitive code.
* Code is more Swift-idiomatic.
* The `localMode` button no longer has the active border, and its color change, when active, is animated.

### DIFF
--- a/macos/Onit/UI/Content/Toolbar.swift
+++ b/macos/Onit/UI/Content/Toolbar.swift
@@ -18,7 +18,7 @@ struct Toolbar: View {
             .padding(.top, 1)
             .padding(.trailing, 8)
             .padding(.leading, 12)
-            .frame(height: 37)
+            .frame(height: 31)
             
             PromptDivider()
         }
--- a/macos/Onit/UI/Content/ToolbarLeft.swift
+++ b/macos/Onit/UI/Content/ToolbarLeft.swift
@@ -11,7 +11,7 @@ struct ToolbarLeft: View {
     @Environment(\.windowState) private var state
     
     var body: some View {
-        HStack(alignment: .center, spacing: 4) {
+        HStack(alignment: .center, spacing: 0) {
             esc
             
             HStack(alignment: .center, spacing: 0) {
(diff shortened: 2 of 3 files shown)


timlenardo:
@lk340 - When I collapse Onit to the smallest width, the 'ESC' text collapses into a '...' Can you make sure it always displays the complete "ESC" word? 

Screenshot 2025-06-24 at 4 55 42 PM

macos/Onit/UI/Content/Toolbar.swift
```diff
@@ -0,0 +1,27 @@
...
+//
+
+import SwiftUI
+
+struct Toolbar: View {
+    var body: some View {
+        VStack(alignment: .leading, spacing: 0) {
+            HStack(alignment: .center) {
+                ToolbarLeft()
+                TetheredToAppView()
```
timlenardo, line 15: This can get removed now!

timlenardo: If you remove this, it fixes the issue I commented about where the ESC button is getting collapsed. But, it introduces some padding issues that should be cleaned up 🧼

timlenardo (changes requested):
Small change requests, otherwise looking good!

timlenardo:
In [d2f2be4](https://github.com/synth-inc/onit/pull/287/commits/d2f2be425e67de174659d18c3dfa1876837489f6), I removed the TetherToAppView (and some commas that make it not compile on my computer - we should discuss this), so you can disregard my prior comments. Now, what's left is that the horizontal padding of the 'ESC' and new chat buttons looks a bit wonky: 

Screenshot 2025-06-24 at 5 07 00 PM

Can you reduce that to match our designs? The "incognito" file in Figma has a design with ESC button in it.

PR #288, opened 2025-06-24
Local tool calling

jrswanson:


macos/Onit/Data/Fetching/Endpoints/Chat/AnthropicChatEndpoint.swift
```diff
@@ -79,23 +108,121 @@ struct AnthropicChatRequest: Codable {
...
         case name
-        case maxUses = "max_uses"
+        case description
+        case input_schema
+        case max_uses
     }
 
     static func search(maxUses: Int) -> AnthropicChatTool {
-        return AnthropicChatTool(type: "web_search_20250305", name: "web_search", maxUses: maxUses)
+        return AnthropicChatTool(type: "web_search_20250305", name: "web_search", max_uses: maxUses)
```
timlenardo, line 124: Best to move this to server config so we can update it without updating the app!

macos/Onit/Data/Tools/CalendarTool.swift
```diff
@@ -0,0 +1,240 @@
...
+    @MainActor
+    static var selectedTools: [String] = ["create_event"]
+
+    @MainActor
+    static var activeTools: [Tool] {
+        return
+            availableTools
+            .filter { selectedTools.contains($0.key) }
+            .map { availableTools[$0.key]! }
+    }
```
timlenardo, line 68: We should filter this for debug & beta builds, or add a toggle in Settings. If we merge this as it's currently written, we will inadvertently ship the calendar tool into prod with the next release, without any UI explanation.

macos/Onit/Info.plist
```diff
@@ -43,6 +43,10 @@
         <string>public.app-category.utilities</string>
         <key>NSApplicationCrashOnExceptions</key>
         <true/>
+        <key>NSCalendarsWriteOnlyAccessUsageDescription</key>
+        <string>Onit would like to automatically create and manage your events through Calendar</string>
+        <key>NSCalendarsFullAccessUsageDescription</key>
+        <string>Onit would like to automatically create and manage your events through Calendar</string>
```
timlenardo, line 49: Let's wrap this in a #if DEBUG | BETA tag, since it's a scary permission to ask for with clear UI in the product. We can remove when we're ready to ship the local tools.

macos/Onit/Onit.entitlements
```diff
@@ -8,6 +8,8 @@
 	<true/>
 	<key>com.apple.security.network.client</key>
 	<true/>
+	<key>com.apple.security.personal-information.calendars</key>
+	<true/>
```
timlenardo, line 12: Same here

macos/Onit/UI/Components/ToolCallView.swift
```diff
@@ -0,0 +1,217 @@
...
+//  ToolCallView.swift
+//  Onit
+//
+//  Created by Jay Swanson on 6/13/25.
+//
+
+import Foundation
+import SwiftUI
+
+struct ToolCallView: View {
```
timlenardo, line 11: This can be a task for @lk340  in the upcoming agent UI sprint, but this class will need to be migrated into LLMStream (and possibly rewritten in JS, since LLMStream is an embedded JS view).

macos/Onit/UI/Panels/State/OnitPanelState+Chat.swift
```diff
@@ -139,6 +139,8 @@ extension OnitPanelState {
                 }
 
                 streamedResponse = ""
+                var functionName = ""
+                var functionArguments = ""
```
timlenardo, line 143: Unless there's a specific reason to use 'function' here, I'd advocate for migrating this (and the params in the various chat endpoints above) to "tool". Here, it'd be "toolName" and "toolArguments".

"function" is clear to me in the context of reviewing this PR, since I know contextually that it must be part of tool use. But if I came across "functionName" in the code, it wouldn't immediately be evident to me that that's synonmyous with a tool call. Let's use one continuous noun, when possible!

macos/Onit/UI/Panels/State/OnitPanelState+Chat.swift
```diff
@@ -250,7 +263,11 @@ extension OnitPanelState {
                             responses: responsesHistory,
                             model: model)
                         for try await response in asyncText {
-                            streamedResponse += response
+                            streamedResponse += response.content ?? ""
+                            if response.functionName != nil {
+                                functionName = response.functionName ?? ""
+                            }
```
timlenardo, line 269: Duplicate nil checks. Can we collapse to one line like the arguments below?

macos/Onit/UI/Prompt/Generated/GeneratedContentView.swift
```diff
@@ -58,6 +64,14 @@ struct GeneratedContentView: View {
...
     }
     
+    private var currentResponse: Response? {
+        guard prompt.generationIndex >= 0,
+              prompt.generationIndex < prompt.sortedResponses.count else {
+            return nil
+        }
+        return prompt.sortedResponses[prompt.generationIndex]
+    }
+
```
timlenardo, line 74: nit: I'd move this to helper to the Prompt object so we can re-use it elsewhere.

macos/Onit/UI/Components/ToolCallView.swift
timlenardo: (I'm otherwise not going to review this file, since the re-write is imminent)

timlenardo (approved):
Okay, I've left some comments below, but otherwise ready to merge. The important things to fix before merging are: 

- Let's be sure to wrap this in #if DEBUG | BETA tags so we don't accidentally release the calendar tool without accompanying UI. 
- Let's migrate to using "tool" everywhere, instead of using "tool" and "function" interchangeably (unless I'm missing something). 

Otherwise, a handful of nits but generally looks good!

macos/Onit/Data/Fetching/Streaming/ChatStreamingEndpointBuilder.swift
```diff
@@ -92,7 +94,8 @@ struct ChatStreamingEndpointBuilder {
         images: [[URL]],
         responses: [String],
         systemMessage: String,
-        userMessages: [String]
+        userMessages: [String],
+        tools: [Tool]
```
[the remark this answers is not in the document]
jrswanson: Good point! We should check

PR #292, opened 2025-06-24
Solution for Environments+Values crash

timlenardo:
Here is one potential solution for the Environment+Values crash that we're seeing in production. 

**Repro steps are as follows:** 
1. Generate a response (my prompt is "write an email about XYZ")
4. Open Gmail, have Gmail foregrounded. 
5. Click Onit's 'copy' button while Gmail is in the foreground. 
6. Repeat this until the crash occurs (usually takes 3 or 4 tries). 

**The issue seems to stem from the following:** 
1. Our main ContentView is in an NSHostingView, which is a legacy MacOS abstraction. 
2. NSHostingView uses async rendering, which performs UI work on a background thread called "DisplayLink".
3. Our main SwiftUI code references an OnitPanelState object, stored in the SwiftUI Environment. When the DisplayLink tries to do UI layout work, it needs to read this OnitPanelState object. 
4. OnitPanelState is marked as @MainActor, so the app crashes when DisplayLink reads it. 

Given that cause, I see a few possible paths to explore: 
1. We could make the windowState optional, and update all of our views to work without a state object. 
2. We could make OnitPanelState not @MainActor isolated.
3. We could explore alternatives to NSHostingView, like SwiftUI's Window or WindowGroup.  

This PR implements option #1. I have confirmed that it works by following the repro steps above. I have a print statement when the windowState gets read and we're not on the main thread. I can see that print statement in the logs, and the app no longer crashes.  I don't like this solution very much. It's annoying to have to unwrap the windowState in every view, and it seems likely that we're introducing bugs. However, it's much better than the current crash. 

Regarding the other solutions, I didn't try option #2, given how closely OnitPanelState is integrated with our UI. It seems like it will be tough, perhaps impossible, to have OnitPanelState off the MainActor. It seems almost sure to introduce bugs.  I haven't tried option #3, but I think it's worth looking into. 

*NOTES* This is marked as DO NOT MERGE because I have not thoroughly tested the code. It is possible that I've introduced bugs in handling the optional windowState. We need to review each change closely and ensure that we do not introduce new crashes when the windowState is optional. Also, I used Cursor heavily when making this, and it seems to have introduced some unrelated code. I'll need to go through and remove that, too.

### DIFF
--- a/macos/Onit/UI/Chat/ChatSystemPromptView.swift
+++ b/macos/Onit/UI/Chat/ChatSystemPromptView.swift
@@ -91,7 +91,7 @@ struct ChatSystemPromptView: View {
                                     .foregroundStyle(.blue300)
                                     .underline()
                                     .onTapGesture {
-                                        state.newChat(shouldSystemPrompt: true)
+                                        state?.newChat(shouldSystemPrompt: true)
                                     }
                             }
                         }
--- a/macos/Onit/UI/Components/TextViewWrapper.swift
+++ b/macos/Onit/UI/Components/TextViewWrapper.swift
@@ -149,6 +149,7 @@ struct TextViewWrapper: NSViewRepresentable {
                     
                     // Prevents errors due to trying to cancel a Task that's already been cancelled.
                     guard !Task.isCancelled else { return }
+                    guard let windowState = parent.windowState else { return }
                     
                     let urls = detectURLs(in: text)
                     
@@ -157,7 +158,7 @@ struct TextViewWrapper: NSViewRepresentable {
                     var textWithoutWebsiteUrls = text
                     
                     for url in urls {
-                        guard let pendingContextList = parent.windowState?.getPendingContextList() else { continue }
+                        let pendingContextList = windowState.getPendingContextList()
                         
                         let urlExists = pendingContextList.contains { context in
                             if case .web(let existingWebsiteUrl, _, _) = context {
@@ -167,7 +168,7 @@ struct TextViewWrapper: NSViewRepresentable {
                         }
                         
                         if !urlExists {
-                            parent.windowState?.addContext(urls: [url])
+                            windowState.addContext(urls: [url])
                             
                             textWithoutWebsiteUrls = removeWebsiteUrlFromText(
                                 text: textWithoutWebsiteUrls,
@@ -222,29 +223,6 @@ struct TextViewWrapper: NSViewRepresentable {
             DispatchQueue.main.async {
                 self.parent.dynamicHeight = height
             }
-        }
-
-        @MainActor func textView(_ textView: NSTextView, shouldInteractWith URL: URL, in characterRange: NSRange, interaction: Any) -> Bool {
-            // Handle URL interactions
-            
-            // Check if windowState is available before accessing
-            guard let windowState = parent.windowState else { return false }
-            
-            let pendingContextList = windowState.getPendingContextList()
-            
-            let urlExists = pendingContextList.contains { context in
-                if case .web(let existingWebsiteUrl, _, _) = context {
-                    return existingWebsiteUrl.absoluteString == URL.absoluteString
-                }
-                return false
-            }
-            
-            if urlExists {
-                return false
-            }
-            
-            windowState.addContext(urls: [URL])
-            return false
         }
     }
 }
--- a/macos/Onit/UI/Components/TextViewWrapper.swift
+++ b/macos/Onit/UI/Components/TextViewWrapper.swift
@@ -107,7 +107,7 @@ struct TextViewWrapper: NSViewRepresentable {
     }
 
     func makeCoordinator() -> Coordinator {
-        return Coordinator(self)
+        Coordinator(self)
     }
 
     class Coordinator: NSObject, NSTextViewDelegate {
@@ -157,7 +157,7 @@ struct TextViewWrapper: NSViewRepresentable {
                     var textWithoutWebsiteUrls = text
                     
                     for url in urls {
-                        let pendingContextList = parent.windowState.getPendingContextList()
+                        guard let pendingContextList = parent.windowState?.getPendingContextList() else { continue }
                         
                         let urlExists = pendingContextList.contains { context in
                             if case .web(let existingWebsiteUrl, _, _) = context {
@@ -167,7 +167,7 @@ struct TextViewWrapper: NSViewRepresentable {
                         }
                         
                         if !urlExists {
-                            parent.windowState.addContext(urls: [url])
+                            parent.windowState?.addContext(urls: [url])
                             
                             textWithoutWebsiteUrls = removeWebsiteUrlFromText(
                                 text: textWithoutWebsiteUrls,
@@ -222,6 +222,29 @@ struct TextViewWrapper: NSViewRepresentable {
             DispatchQueue.main.async {
                 self.parent.dynamicHeight = height
             }
+        }
+
+        @MainActor func textView(_ textView: NSTextView, shouldInteractWith URL: URL, in characterRange: NSRange, interaction: Any) -> Bool {
+            // Handle URL interactions
+            
+            // Check if windowState is available before accessing
+            guard let windowState = parent.windowState else { return false }
+            
+            let pendingContextList = windowState.getPendingContextList()
+            
+            let urlExists = pendingContextList.contains { context in
+                if case .web(let existingWebsiteUrl, _, _) = context {
+                    return existingWebsiteUrl.absoluteString == URL.absoluteString
+                }
+                return false
+            }
+            
+            if urlExists {
+                return false
+            }
+            
+            windowState.addContext(urls: [URL])
+            return false
         }
     }
 }
--- a/macos/Onit/UI/Content/TetheredButton.swift
+++ b/macos/Onit/UI/Content/TetheredButton.swift
@@ -21,13 +21,14 @@ struct TetheredButton: View {
     }
     
     private var spacerHeight: CGFloat? {
-        guard let relativePosition = state?.tetheredButtonYRelativePosition else { return nil }
+        guard let state = state else { return nil }
+        guard let relativePosition = state.tetheredButtonYRelativePosition else { return nil }
         
         let windowHeight: CGFloat
-        if let trackedWindow = state?.trackedWindow,
+        if let trackedWindow = state.trackedWindow,
            let frame = trackedWindow.element.getFrame(convertedToGlobalCoordinateSpace: true) {
             windowHeight = frame.height
-        } else if let trackedScreen = state?.trackedScreen {
+        } else if let trackedScreen = state.trackedScreen {
             windowHeight = trackedScreen.visibleFrame.height
         } else {
             return nil
@@ -72,11 +73,6 @@ struct TetheredButton: View {
             }
             .tooltip(prompt: fitActiveWindowPrompt)
             Spacer()
-        }
-        .onAppear {
-            if let defaultEnvironmentSource = state?.defaultEnvironmentSource {
-                AnalyticsManager.Technical.defaultWindowState(source: defaultEnvironmentSource)
-            }
         }
         .edgesIgnoringSafeArea(.top)
         .frame(maxHeight: .infinity, alignment: .top)
--- a/macos/Onit/UI/Content/TetheredButton.swift
+++ b/macos/Onit/UI/Content/TetheredButton.swift
@@ -21,13 +21,13 @@ struct TetheredButton: View {
     }
     
     private var spacerHeight: CGFloat? {
-        guard let relativePosition = state.tetheredButtonYRelativePosition else { return nil }
+        guard let relativePosition = state?.tetheredButtonYRelativePosition else { return nil }
         
         let windowHeight: CGFloat
-        if let trackedWindow = state.trackedWindow,
+        if let trackedWindow = state?.trackedWindow,
            let frame = trackedWindow.element.getFrame(convertedToGlobalCoordinateSpace: true) {
             windowHeight = frame.height
-        } else if let trackedScreen = state.trackedScreen {
+        } else if let trackedScreen = state?.trackedScreen {
             windowHeight = trackedScreen.visibleFrame.height
         } else {
             return nil
@@ -41,7 +41,7 @@ struct TetheredButton: View {
             return .degrees(0)
         }
 
-        if state.trackedScreen != nil || state.panel == nil || state.panel?.resizedApplication == true {
+        if state?.trackedScreen != nil || state?.panel == nil || state?.panel?.resizedApplication == true {
             return .degrees(0)
         } else {
             return .degrees(180)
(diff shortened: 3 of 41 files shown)


macos/Onit/UI/Components/TextViewWrapper.swift
```diff
@@ -223,6 +223,29 @@ struct TextViewWrapper: NSViewRepresentable {
...
+                    return existingWebsiteUrl.absoluteString == URL.absoluteString
+                }
+                return false
+            }
+            
+            if urlExists {
+                return false
+            }
+            
+            windowState.addContext(urls: [URL])
```
timlenardo, line 246: TODO: do we need this?

macos/Onit/UI/Prompt/Files/ContextMenuWindows.swift
```diff
@@ -325,4 +325,49 @@ extension ContextMenuWindows {
...
+    }
+    
+    private func getCurrentContext() -> Context? {
+        return windowState?.getPendingContextList().first { context in
+            if case .web = context {
+                return false
+            }
+            return true
+        }
+    }
```
timlenardo, line 340: I think we can remove all of this

macos/Onit/UI/Prompt/Files/WebContextItem.swift
```diff
@@ -104,4 +106,25 @@ extension WebContextItem {
...
+        
+        let pendingContextList = windowState.getPendingContextList()
+        
+        return pendingContextList.contains { context in
+            if case .web(let existingWebsiteUrl, _, _) = context {
+                return existingWebsiteUrl.absoluteString == websiteUrl.absoluteString
+            }
+            return false
+        }
+    }
```
timlenardo, line 129: TOD: remove, shouldn't be adding methos, just making state optional

macos/Onit/UI/Prompt/MicrophoneButton.swift
```diff
@@ -310,4 +314,32 @@ extension MicrophoneButton {
...
+        let cursorPosition = windowState.pendingInstructionCursorPosition
+        let spaces = "    " // 4 spaces for indentation
+        
+        let targetIndex = windowState.pendingInstruction.index(
+            windowState.pendingInstruction.startIndex,
+            offsetBy: min(max(0, cursorPosition), windowState.pendingInstruction.count)
+        )
+        windowState.pendingInstruction.insert(contentsOf: spaces, at: targetIndex)
+    }
 }
```
timlenardo, line 345: TODO: Should also remove this- no new functions, just making state optional

macos/Onit/UI/Prompt/PromptCore.swift
```diff
@@ -85,7 +85,7 @@ struct PromptCore: View {
         }
         .background {
             if !isEditing {
-                if !showSlashMenu && !windowState.showContextMenu && !windowState.isTyping {
+                if !showSlashMenu && (windowState?.showContextMenu != true) && (windowState?.isTyping != true) {
```
timlenardo, line 88: How does this get evaluated if windowState is nil?

macos/Onit/UI/Prompt/PromptCore.swift
```diff
@@ -136,39 +136,40 @@ struct PromptCore: View {
...
+            .onChange(of: windowState.textFocusTrigger) { _, _ in
+                isFocused = true
+            }
+            .onChange(of: showingAlert) { _, new in
+                isFocused = !new
+            }
+            .disabled(showingAlert)
+            .allowsHitTesting(!showingAlert)
+        } else {
+            Text("Loading...")
```
Niduank, line 168: Like all the files that add an `EmptyView()` when windowState is nil, we should do the same.
Or better, display an error view, because if this happens, the app will be in a broken state.

macos/Onit/UI/Prompt/PromptCore.swift
```diff
@@ -242,7 +245,8 @@ extension PromptCore {
 
 extension PromptCore {
     private var shouldIndicateDisabled: Bool {
-        !windowState.websiteUrlsScrapeQueue.isEmpty || !windowState.windowContextTasks.isEmpty
+        guard let windowState = windowState else { return false }
```
Niduank, line 248: If there is no windowState, `shouldIndicateDisabled = true`
`guard let windowState = windowState else { return true }`

macos/Onit/UI/Prompt/PromptCore.swift
```diff
@@ -257,22 +261,26 @@ extension PromptCore {
     }
     
     private var placeholderText: String {
+        guard let windowState = windowState else {
+            return "Loading..."
```
Niduank, line 265: Replace "Loading..." by an error message.

macos/Onit/UI/Prompt/SystemPrompt/SystemPromptDetailView.swift
```diff
@@ -114,7 +115,7 @@ struct SystemPromptDetailView: View {
             
             Spacer()
             
-            if windowState.systemPromptId != SystemPrompt.outputOnly.id {
+            if windowState?.systemPromptId != SystemPrompt.outputOnly.id {
```
Niduank, line 118: If windowState is nil, we shouldn't show the `Button`
```swift
if let windowState = windowState, windowState.systemPromptId != SystemPrompt.outputOnly.id {
```

macos/Onit/UI/Prompt/SystemPrompt/SystemPromptSelectionView.swift
```diff
@@ -22,13 +22,13 @@ struct SystemPromptSelectionView: View {
         self._showNewPrompt = showNewPrompt
     }
     
-    private var suggestedPrompts: [SystemPrompt] {
-        state.promptSuggestionService?.suggestedPrompts ?? []
+    private var systemPrompts: [SystemPrompt] {
```
Niduank, line 25: Can you rename the variable as it was before : `suggestedPrompts`

macos/Onit/UI/Windows/ContextPickerView.swift
```diff
@@ -61,7 +61,7 @@ extension ContextPickerView {
         } else if !autoContextFromCurrentWindow {
             return "Click to enable"
         } else {
-            if let foregroundWindow = windowState.foregroundWindow {
+            if let windowState = windowState, let foregroundWindow = windowState.foregroundWindow {
                 return WindowHelpers.getWindowName(window: foregroundWindow.element)
```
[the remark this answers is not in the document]
Niduank: I agree

macos/Onit/UI/Prompt/Files/ContextMenuWindows.swift
```diff
@@ -325,4 +325,49 @@ extension ContextMenuWindows {
         isCapturingWindows = false
         return orderCapturedWindowsByMostRecent(capturedOpenWindows)
     }
+    
+    private func handleAddFiles() {
+        windowState?.showFileImporter = true
+    }
```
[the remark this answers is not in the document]
timlenardo: It looks like this function has been removed in main. I won't make any changes to avoid unnecessary conflicts.

macos/Onit/UI/Components/TextViewWrapper.swift
timlenardo: Yeah, I'll remove. We shouldn't make any changes here that aren't related to making the windowState optional.

macos/Onit/UI/Components/TextViewWrapper.swift
```diff
@@ -157,7 +157,7 @@ struct TextViewWrapper: NSViewRepresentable {
                     var textWithoutWebsiteUrls = text
```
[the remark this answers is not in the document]
timlenardo: Agreed, done!

macos/Onit/UI/Content/TetheredButton.swift
```diff
@@ -75,7 +75,7 @@ struct TetheredButton: View {
             Spacer()
         }
         .onAppear {
-            if let defaultEnvironmentSource = state.defaultEnvironmentSource {
+            if let defaultEnvironmentSource = state?.defaultEnvironmentSource {
```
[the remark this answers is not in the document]
timlenardo: Will do!

macos/Onit/UI/Environment/Environment+Values.swift
```diff
@@ -60,8 +31,15 @@ extension EnvironmentValues {
         set { self[OnitAppStateKey.self] = newValue }
     }
     
-    var windowState: OnitPanelState {
-        get { self[OnitPanelStateKey.self] }
+    var windowState: OnitPanelState? {
+        get { 
+            if Thread.isMainThread {
```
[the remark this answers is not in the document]
timlenardo: Okay, agreed.

macos/Onit/UI/History/HistoryRowView.swift
```diff
@@ -2,66 +2,65 @@
...
-//  Created by Benjamin Sage on 11/4/24.
+//  Created by Benjamin Sage on 11/3/24.
 //
 
 import SwiftUI
 
 struct HistoryRowView: View {
     @Environment(\.windowState) private var windowState
+    let chat: Chat
+    let index: Int
```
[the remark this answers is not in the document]
timlenardo: Okay, yeah, we shouldn't make any changes other than making the windowState optional in this PR, so I will revert this view

macos/Onit/UI/History/HistoryTitle.swift
```diff
@@ -11,23 +11,30 @@ struct HistoryTitle: View {
...
+                }
+                .buttonStyle(BorderlessButtonStyle())
             }
-            .buttonStyle(BorderlessButtonStyle())
+            .padding(.horizontal, 16)
+            .padding(.vertical, 12)
+        } else {
+            Text("History")
+                .appFont(.medium16)
+                .foregroundStyle(.white)
```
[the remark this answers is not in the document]
timlenardo: Skipping as we're removing this fallback state.

macos/Onit/UI/Prompt/ChatsView.swift
```diff
@@ -12,13 +12,18 @@ struct ChatsView: View {
     @Environment(\.windowState) private var state
     
     var body: some View {
-        VStack(spacing: 0) {
-            ForEach(state.currentPrompts ?? []) { prompt in
```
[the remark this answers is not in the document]
timlenardo: Agreed, that's cleaner. Updating!

macos/Onit/UI/Prompt/Dialogs/SetUpDialogs.swift
```diff
@@ -46,47 +46,51 @@ struct SetUpDialogs: View {
...
-            guard state.panel?.isVisible == true else {
-                state.setUpHeight = 0
+            guard state?.panel?.isVisible == true else {
+                state?.setUpHeight = 0
                 return
             }
             
-            state.setUpHeight = newHeight
+            state?.setUpHeight = newHeight
         }
```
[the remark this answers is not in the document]
timlenardo: Okay, will do!

macos/Onit/UI/Prompt/Files/ContextMenu.swift
```diff
@@ -14,38 +14,45 @@ struct ContextMenu: View {
...
+        .background(Color.gray800)
+        .cornerRadius(12)
+        .onTapGesture {
+            windowState?.showContextMenu = false
+        }
+        .onChange(of: showBrowserTabsContextMenu) { _, new in
+            if new {
+                windowState?.showContextMenuBrowserTabs = true
+            }
         }
```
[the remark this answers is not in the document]
timlenardo: Skipping since we've reverted.

macos/Onit/UI/Prompt/Files/ContextMenuWindowButton.swift
```diff
@@ -39,23 +39,28 @@ struct ContextMenuWindowButton: View {
     }
     
     var body: some View {
-        TextButton(
```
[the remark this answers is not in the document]
timlenardo: Good catch, removing this!

macos/Onit/UI/Prompt/Files/ContextMenuWindows.swift
```diff
@@ -325,4 +325,49 @@ extension ContextMenuWindows {
...
+        guard let windowState = windowState else { return [] }
+        
+        let windowContextTaskWindowHashes = Set(windowState.windowContextTasks.keys)
+        
+        var windowsToFetch: [TrackedWindow] = []
+        
+        // TODO: Implement the logic for fetching foregrounded windows
+        
+        return windowsToFetch
+    }
```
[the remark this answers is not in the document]
timlenardo: Skipping, as these functions were unneeded AI slop.

macos/Onit/UI/Prompt/PromptCore.swift
timlenardo: I'm going to do an EmptyView because I don't think we should ever be in this state, except for a few microseconds when context switching between apps. We can spend the time designing and implementing an error state if we determine that it is needed.

macos/Onit/UI/Prompt/PromptCore.swift
timlenardo: nice catch!

macos/Onit/UI/Prompt/PromptCore.swift
timlenardo: I am going to make this an empty string, because I believe we can end up in this state very briefly (like a few ms) when switching between applications. If we show an error/loading message, the placeholder may flicker. Better to do an empty string, I think, and then only the correct message will appear, once the state is loaded.

macos/Onit/UI/Prompt/SystemPrompt/SystemPromptSelectionView.swift
timlenardo: Yup, not sure why Cursor makes these unrelated changes 😅

macos/Onit/UI/Chat/ChatSystemPromptView.swift
```diff
@@ -91,7 +91,7 @@ struct ChatSystemPromptView: View {
                                     .foregroundStyle(.blue300)
                                     .underline()
                                     .onTapGesture {
-                                        state.newChat(shouldSystemPrompt: true)
+                                        state?.newChat(shouldSystemPrompt: true)
```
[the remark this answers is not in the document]
Niduank: I guess I was going to comment “The action won’t do anything,” but that’s actually expected and shouldn’t happen anyway.

macos/Onit/UI/Prompt/PromptCore.swift
```diff
@@ -136,7 +136,7 @@ struct PromptCore: View {
 extension PromptCore {
     @ViewBuilder
     private var textField: some View {
-        if let windowState = windowState {
+        if let windowState = windowState, false {
```
Niduank, line 139: This hard-coded `false` avoid to display the text input.
 
Screenshot 2025-07-08 at 12 07 06

Niduank: I fixed it. Now everything looks good 👍 You can merge it

PR #294, opened 2025-06-25
Resize all windows in Pinned mode, not just tracked ones

timlenardo:
This PR fixes a small but persistent issue: other windows weren’t always being resized when closing the Onit panel in Pinned mode. This happened frequently for me while developing Onit. Here’s the scenario: when I open Onit to debug, it shrinks some windows. If I quit Onit by pressing “stop” in Xcode, those windows stay shrunken. The next time I launch Onit and then close it, the other windows don’t get resized back to normal, because they’re not in ‘targetInitialFrames.’ That’s because Onit didn’t resize them this time—they were already resized by the previous instance.

This change improves the logic to match how we handle window resizing: now, Onit scans all windows from all other active processes and resizes any that border Onit, regardless of whether the current instance resized them.

Additionally, I refactored some duplicate code: the logic to find all other process IDs and windows is now moved into helper methods in WindowHelper.

### DIFF
--- a/macos/Onit/Helpers/WindowHelpers.swift
+++ b/macos/Onit/Helpers/WindowHelpers.swift
@@ -44,4 +44,26 @@ struct WindowHelpers {
             return nil
         }
     }
+    
+    /// Returns all PIDs of regular running applications excluding the current app (Onit)
+    static func getAllOtherAppPids() -> [pid_t] {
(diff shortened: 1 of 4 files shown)


jrswanson (approved):
Looks good! Pulling these into WindowHelpers cleans things up a lot and this bug was really annoying to develop with.

PR #295, opened 2025-06-25
Use safe generation index when deleting responses

timlenardo:
This offers a solution to an 'out of bounds' error that emerged when we added the "stopGeneration" button in v2.5. 

The issue is caused by the following: 
- When we stop a generation, if it's the first 'instruction' for the current prompt, we delete the prompt. 
- We update the underlying "Chat" to remove the prompt and also delete the Prompt itself. 
- Finally we reset the generationIndex, removing 1 to account for the deletion.
- Many views are listening for changes in Chat and Prompt. 
- We have race conditions, where the a view can update before the generationIndex is set. 
- Specifically, this was causing out-of-bounds error in 'textToRead()' function of GeneratedContentView. We were reading from 'sortedResponses' after a response had been removed, but before the generationIndex had been updated. 

In this PR, I  introduce a "safeGenerationIndex", a computed property that is always within the bounds of the sortedResponses array. I also added a "validateGenerationIndex" function, which is always called when we update the responses array. Finally, I added safe helper methods for removing responses that automatically update all of this. From local testing, this appears to resolve the crash and also provides us with helper methods for deleting responses in the future.

### DIFF
--- a/macos/Onit/Data/Persistence/Prompt.swift
+++ b/macos/Onit/Data/Persistence/Prompt.swift
@@ -16,7 +16,11 @@ import SwiftData
     var contextList: [Context] = []
 
     //    @Relationship(deleteRule: .cascade, inverse: \Response.prompt)
-    var responses: [Response] = []
+    var responses: [Response] = [] {
+        didSet {
+            validateGenerationIndex()
+        }
+    }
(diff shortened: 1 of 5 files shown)


jrswanson (approved):
LGTM

PR #298, opened 2025-06-26
Search polishing

jrswanson:
We just merged three outstanding PR's related to web search (#291, #290, #289). This PR puts these features and updates in a better user facing place.
- Added support for OpenAI and Anthropic's search tools for users with valid provider tokens
- Now Onit will dynamically query the search providers available on the server to better select a web search strategy
- Added a Tavily costing saving mode for when users prefer to use Tavily over our credit based system
- Added an option to enable/disable search when local mode is active to maintain privacy

### DIFF
--- a/macos/Onit/Data/Fetching/Endpoints/ChatStreaming/AnthropicChatStreamingEndpoint.swift
+++ b/macos/Onit/Data/Fetching/Endpoints/ChatStreaming/AnthropicChatStreamingEndpoint.swift
@@ -26,7 +26,7 @@ struct AnthropicChatStreamingEndpoint: StreamingEndpoint {
     var requestBody: AnthropicChatRequest? {
         var tools: [AnthropicChatTool] = []
         if includeSearch == true {
-            tools.append(AnthropicChatTool(type: "web_search_20250305", name: "web_search", maxUses: 5))
+            tools.append(AnthropicChatTool.search(maxUses: 5))
         }
         return AnthropicChatRequest(
             model: model,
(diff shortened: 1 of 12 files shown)


macos/Onit/Data/Fetching/Endpoints/ChatStreaming/AnthropicChatStreamingEndpoint.swift
```diff
@@ -41,14 +47,14 @@ struct AnthropicChatStreamingEndpoint: StreamingEndpoint {
...
-        let jsonString = String(eventString[dataStart...]).trimmingCharacters(in: .whitespacesAndNewlines)
         
-        if let data = jsonString.data(using: .utf8) {
+        if let data = event.data?.data(using: .utf8) {
             let response = try JSONDecoder().decode(Response.self, from: data)
             
+            if response.contentBlock?.type == "server_tool_use" {
+                return "\n\n...\n\n"
+            }
+
```
timlenardo, line 57: This is interesting. When does this happen? I've been using it with Anthropic and haven't seen any '\n\n...\n\n's in the responses so far.

jrswanson: This happens when there's a bit of text before "Certainly I'll get started right away!" then a search tool call, and then a response

jrswanson: It's just a placeholder till we have our own spinner/shimmer

PR #299, opened 2025-06-26
Feature: Quick Edit (command K)

Niduank:


timlenardo (approved):
There is a lot of work left to be done on this feature, but since it's set to only be available in debug and Beta builds, we can merge it. I am going to squash and rebase these commits and then merge into main! Then, I will start a series of follow-up PRs in different branches.

Issue #283, opened 2025-06-21
No longer local-first


alanpearce:
I'm confronted with a login/sign-up screen when I want to use Onit now. This makes the description of "local-first" misleading. I would like it if I did not have to create an account to use my own API keys or local models. Alternatively I would suggest removing the "local-first" descriptor from the website or adding a small text to mention that an Onit account is nevertheless required.

timlenardo:
Yup- I noticed this too when I was flying last weekend. That wasn't the intent; it just got missed because we all use it predominantly with internet access. "Support offline mode" is on our priority list, but all our engineers are heads down on other major new features right now, so it may be a few weeks before we get to this. 

@alanpearce @andyylin @mcwq123 , if any of you are developers, feel free to submit a PR!

PR #305, opened 2025-07-01
Local-First Support ("offline mode")

lk340:
### Task

* This update addresses [recent user feedback](https://github.com/synth-inc/onit/issues/283), requesting there to be explicit local-first support.
* The implication behind local-first support is two-fold:
    1. User should not have to create an account to access the app if they have provided their own local model.
    2. User should not have to create an account to access the app if they have provided at least one provider API key.

### Notes

* Auth flow is now only fixed to the UI when the user is not logged in AND either hasn't provided their own local model or provider API key.
* After providing a local model or provider API key, the auth flow automatically disappears but is still accessible through the settings "General" tab.
* The auth flow has now been updated to show a CTA for when the user wants to provide their own models.  If they have provided their own local model or provider API key, this CTA disappears and a "close" button appears at the bottom of the auth flow to allow for continuous use of the app while logged out.
* If the user has provided a local model but NOT a provider API key, they will not have access to any remote models and will, instead, be shown a CTA in the model selection view to either log in or add their own provider API key to access remote models.
    * This means that, in this scenario, the model mode toggle shortcut is disabled, and local mode is strictly enforced.
* If the user is not logged in but but has added their own provider API keys, the local model restriction will be lifted, and they will only have access to remote models whose provider API keys were provided.

### DIFF
--- a/macos/Onit/AppState.swift
+++ b/macos/Onit/AppState.swift
@@ -80,7 +80,9 @@ class AppState: NSObject, SPUUpdaterDelegate {
     var subscriptionPlanError: String = ""
     
     // Network Monitoring
+    var networkMonitor = NetworkMonitor.shared
     var isOnline: Bool = NetworkMonitor.shared.isOnline
+    
     private var cancellables = Set<AnyCancellable>()
 
     // MARK: - Initializer
@@ -95,10 +97,26 @@ class AppState: NSObject, SPUUpdaterDelegate {
             userDriverDelegate: nil
         )
         
-        NetworkMonitor.shared.$isOnline
+        networkMonitor.$isOnline
             .receive(on: DispatchQueue.main)
-            .sink{ [weak self] online in
-                self?.isOnline = online
+            .sink{ [weak self] isOnline in
+                if let self = self {
+                    self.isOnline = isOnline
+                    
+                    if !self.isOnline {
+                        Defaults[.mode] = .local
+                    } else {
+                        if !userLoggedIn {
+                            if self.hasUserAPITokens {
+                                Defaults[.mode] = .remote
+                            } else if hasLocalModels {
+                                Defaults[.mode] = .local
+                            }
+                        } else {
+                            Defaults[.mode] = .remote
+                        }
+                    }
+                }
             }
             .store(in: &cancellables)
 
@@ -577,6 +595,14 @@ class AppState: NSObject, SPUUpdaterDelegate {
     
     var canAccessRemoteModels: Bool {
         (self.userLoggedIn || self.hasUserAPITokens) && self.isOnline
+    }
+    
+    @ObservableDefault(.availableLocalModels)
+    @ObservationIgnored
+    var availableLocalModels: [String]
+    
+    var hasLocalModels: Bool {
+        !availableLocalModels.isEmpty
     }
 }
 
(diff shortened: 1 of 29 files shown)


timlenardo (changes requested):
Hey @lk340 - thanks for taking this on! I'm testing now and observed some issues: 

When I launch the app without Wifi, the PromptCore highlight correctly indicates that we're in local mode, but the 'local mode' button is not green:
Screenshot 2025-07-03 at 1 07 53 PM

In general, the app is not responsive to changes in WiFi conditions. For example, if I re-enable Wifi, while the app is running, the local/remote mode toggle is not interactable and I can't switch back into remote mode. I can only re-enable remote mode by quitting and re-opening the app. I think we want to set this up as a hook on the wifi status, instead of only checking when the app launches:
Screenshot 2025-07-03 at 1 08 19 PM

Also, having added custom tokens doesn't get us much in remote mode: they still can't access those servers (OpenAI, etc) without WIFI. Only Ollama (or locally hosted custom endpoints) will work. So I don't think our logic here (and related copy) is quite right. Instead of "have local models or API key" is should just be "have local models?". 

Screenshot 2025-07-03 at 12 59 25 PM

timlenardo:
Hi @lk340, I spent some time testing this afternoon and encountered several issues.  This is going to be a tough PR to get right, since it introduces a lot of new states:

- logged in vs. logged out
- Has internet vs. no internet
- Has remote api key vs. no remote API keys
- Has Ollama vs. Ollama not running.

Plus some edge cases that we're handling, like:
- Ollama set up before, but not current running: 
- Remote API keys previously working, but not anymore. 

Across the app, we'll need to audit all elements that listen to ANY of these states, and make sure they are handling every possible case above. Then, for all future PRs, we'll need to do the same checks. Every new element that listens to any of the above, needs to handle all possible cases. 

Here are some of the things I saw, in no particular order: 


1.  After I log out, in settings, I still have access to all the same remote models that I had when I was logged in. 
- We should clear these on logout, but also…
- We should save those settings on a per-account basis, so if I log out and then log back in, it remembers which models I selected. 

2. After I log out, I still have items in History, even though I’ve logged out. How is this working? Are those items I searched when I was testing it in logged out mode before?

3. Somehow I end up logged back in after restarting the app, even though I had logged out. 

4. Somehow, when I get past the login page by adding the remote model key, I end up in local mode. I think because I was in local mode before? We should use the local/remote mode based on which method was used to advance past the login screen. 

5. If I’m using it without a network, I see an error message in the remote model selection dropdown that says, “Sign up, login, or add your own API key…”
	-  Instead, this should say “no network connection.” 

6. When we are showing that same message “sign up, log in, etc.” to users that do have internet connection, let’s add a clickable button to login that opens settings to the General page, in addition to the button that opens the models page to add API tokens. We want to encourage people to log in. 

7. When I am logged in and don’t have Ollama running but then turn off internet connection…
	- The app toggles into local mode, but I don’t have local models active. 
	-  “Choose Model” dropdown just shows an error triangle for “local models’ section. No CTA to set up Ollama or error message.
	-  I think it’s okay to go into remote mode, but it we should have a CTA in the local models selection section encouraging users to set up Ollama to use in local mode. 
	-  When I re-enable internet, it doesn’t go back into remote mode. 

I'm sure there will be more edge cases to handle, but I wanted to get these on your plate so that you can start working through them. Thanks in advance, I know this is a very tricky feature to get right 🙏

PR #306, opened 2025-07-02
Fix: Flashing Window Context Errors

lk340:
### Task

* For large pieces of text, the window context tag's tooltip was either failing (flashing without properly staying on the UI) or bleeding onto an external monitor.
* The tooltip manager is less than ideal for handling large pieces of text, especially since we would want them to be in scrollable containers for better readability.

### Solution
* Update `ContextTag` to have an optional popover element with editable text color and dynamic scrollview.
* This allows us to move large text, like error messages, safely into the `ContextTag`. Error message are now shown in a scrollable popover with `.red` text color.

### DIFF
--- a/macos/Onit/UI/Content/ExternalTetheredButtonStyle.swift
+++ b/macos/Onit/UI/Content/ExternalTetheredButtonStyle.swift
@@ -138,7 +138,7 @@ struct ExternalTetheredButtonStyle: ButtonStyle {
     }
 
     private func handleHover(_ hovering: Bool) {
-        self.hovering = hovering
+        self.isHovering = hovering
         
         TooltipHelpers.setTooltipOnHover(
             isHovering: hovering,


macos/Onit/General/Styles/ViewStyle.swift
```diff
@@ -98,8 +98,8 @@ extension View {
             .underline(underline)
     }
     
-    func truncateText() -> some View {
-        self.lineLimit(1).truncationMode(.tail)
+    func truncateText(lineLimit: Int = 1) -> some View {
+        self.lineLimit(lineLimit).truncationMode(.tail)
```
timlenardo, line 102: I don't think we're using this anywhere?

macos/Onit/UI/Components/TextButton.swift
```diff
@@ -28,14 +28,14 @@ struct TextButton<Child: View>: View {
     private let fontWeight: Font.Weight
     private let fontColor: Color
     
+    private let tooltipShortcut: Tooltip.Shortcut
+    private let tooltipPrompt: String?
+
```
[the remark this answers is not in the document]
lk340: No, this was not AI's doing. I did this because I felt that it'd make it more readable in the long run to show the proper default tooltip value for an empty state. But yes, you're absolutely right in that it makes the review unnecessarily less readable! It's a bit long, but I DM'd you my reasons! 🙇

macos/Onit/UI/Components/TooltipView.swift
```diff
@@ -8,47 +8,104 @@
...
+            .frame(minHeight: 78)
+        }
+    }
+    
+    // MARK: - Child Components
+    
+    @ViewBuilder
+    var text: some View {
+        let textSubstring: [Substring] = tooltip.prompt.split(separator: "\n")
+        let textFirstLine = textSubstring[0]
```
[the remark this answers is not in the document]
lk340: Worked like a charm! 🕺

macos/Onit/UI/Tooltip/TooltipManager.swift
```diff
@@ -24,14 +24,21 @@ class TooltipManager {
...
-        delayEnd: Double = 0.2
+        delayEnd: Double = 0.2,
+        ignoreMouseEvents: Bool = false
     ) {
         tooltipTask?.cancel()
+        
+        /// Persists ignoring of mouse events across different tooltip states.
+        if ignoreMouseEvents {
+            tooltipWindow?.ignoresMouseEvents = true
+        }
```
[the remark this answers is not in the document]
lk340: Yes, it would be my pleasure! So this is actually tied to the key reason as to why we were getting the flashing error when hovering over the context tags. When the tooltip view gets big enough, it can overlap with the mouse cursor. For the cases where we're displaying the TooltipView with the `onHover` view modifier, this causes the `isHovering` closure to "flash" between `true` and `false` indefinitely. `ignoresMouseEvents` allows us to bypass that by making tooltips and the cursor non-interactive, which is fine because a tooltip is just meant to display some text.

An important clarification to make here is that we're *only* ever setting `ignoresMouseEvents` to `true` when manually setting tooltips via `TooltipHelpers.setTooltipOnHover` (the helper function itself defaults to using `false` for `ignoresMouseEvents`). This applies to `ContextTag`, `IconButton`, `TextButton`, and now `ExternalTetheredButtonStyle`.

This property also defaults to `false` in `TooltipManager` to make this update backwards-compatible!

For example, views that directly invoke the `.tooltip` modifier do not set their `ignoresMouseEvents` to `true`! This can be seen in `TooltipModifier` -> `HoverableButtonStyle`.

> Are there situations where we'd want it to be false?

To be honest, as it stands, no. Ideally, we should be setting it to `true` if we want to permanently avoid these tooltip-cursor overlap issues. The reasons as to why I didn't do this is because:

1. Tooltips are being used in two different ways: through the tooltip helper and by directly invoking the `.tooltip` view modifier. It's defaulting to `false` and remains an optional value to keep backwards-compatibility with the latter.
2. Keeping the above in mind, making this an optional value allows us to modify the tooltip view in the future to be interactive, if we'd ever want for it to do that.

macos/Onit/UI/Components/ContextTag.swift
```diff
@@ -38,7 +38,7 @@ struct ContextTag: View {
         iconBundleURL: URL? = nil,
         iconView: (any View)? = nil,
         caption: String? = nil,
-        tooltip: String? = nil,
+        tooltipPrompt: String? = nil,
```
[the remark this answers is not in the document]
timlenardo: Since we agree that calling it 'prompt' is moving in the wrong direction, let's not rename it in this PR (i.e., undo this change and leave it as 'tooltip', as it originally was). Then, in a separate PR later, we can replace 'tooltipPrompt' elsewhere with 'tooltipText' or a similar term.

macos/Onit/UI/Components/TextButton.swift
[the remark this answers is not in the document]
timlenardo: Oh, I just meant moving these two lines from lines 36-38 to lines 31-33. Why did you move it?

macos/Onit/UI/Components/TooltipView.swift
```diff
@@ -8,42 +8,107 @@
...
+    
+    private func findValidSubstring(_ substrings: [String]) -> String {
+        return substrings
+            .first { !$0.trimmingCharacters(in: .whitespaces).isEmpty }?
+            .trimmingCharacters(in: .whitespaces)
+        ?? ""
+    }
+    
+    @ViewBuilder
+    var text: some View {
```
timlenardo, line 38: I know we have this pattern elsewhere, but I'd prefer to name views as views (i.e. `var textView` instead of `var text`). When I see something with a name like 'text', 'fixedText' or 'truncatedText' in code, I generally assume those would be Strings. So, I find the main 'body' confusing here.

macos/Onit/UI/Components/ContextTag.swift
```diff
@@ -64,6 +64,7 @@ struct ContextTag: View {
     @State private var isHoveredBody: Bool = false
     @State private var isPressedBody: Bool = false
     @State private var isHoveredRemove: Bool = false
+    @State private var isPopoverOpen: Bool = false
```
timlenardo, line 67: Remove

macos/Onit/UI/Components/ContextTag.swift
```diff
@@ -196,3 +189,11 @@ extension ContextTag {
...
     }
 }
+
+// MARK: - Private Functions
+
+extension ContextTag {
+    private func togglePopover() {
+        isPopoverOpen.toggle()
+    }
+}
```
timlenardo, line 199: Unused, remove

macos/Onit/UI/Components/IconButton.swift
```diff
@@ -19,8 +19,8 @@ struct IconButton: View {
     private let cornerRadius: CGFloat
     private let activeBorderColor: Color
     
+    private let tooltipShortcut: Tooltip.Shortcut
```
timlenardo, line 22: undo shuffle here (this can go back to line 23)

macos/Onit/UI/Components/IconButton.swift
```diff
@@ -36,8 +36,8 @@ struct IconButton: View {
         cornerRadius: CGFloat = ToolbarButtonStyle.cornerRadius,
         activeBorderColor: Color = .gray500,
         
+        tooltipShortcut: Tooltip.Shortcut = .none,
```
timlenardo, line 39: undo shuffle here (this can go back to line 40)

macos/Onit/UI/Components/IconButton.swift
```diff
@@ -53,8 +53,8 @@ struct IconButton: View {
         self.cornerRadius = cornerRadius
         self.activeBorderColor = activeBorderColor
         
-        self.tooltipPrompt = tooltipPrompt
         self.tooltipShortcut = tooltipShortcut
+        self.tooltipPrompt = tooltipPrompt
```
timlenardo, line 57: undo shuffle here (this can go back to line 56)

macos/Onit/UI/Components/TextButton.swift
```diff
@@ -28,14 +28,14 @@ struct TextButton<Child: View>: View {
     private let fontWeight: Font.Weight
     private let fontColor: Color
     
+    private let tooltipShortcut: Tooltip.Shortcut
+    private let tooltipPrompt: String?
```
timlenardo, line 32: undo shuffle here (this can go back to line 36-37)

macos/Onit/UI/Components/TextButton.swift
```diff
@@ -91,14 +91,14 @@ struct TextButton<Child: View>: View {
         self.fontWeight = fontWeight
         self.fontColor = fontColor
         
+        self.tooltipShortcut = tooltipShortcut
+        self.tooltipPrompt = tooltipPrompt
```
timlenardo, line 95: undo shuffle here (this can go back to line 99-100)

macos/Onit/UI/Content/ToolbarLeft.swift
```diff
@@ -28,9 +28,9 @@ struct ToolbarLeft: View {
             cornerRadius: ToolbarButtonStyle.cornerRadius,
             fontSize: 13,
             fontColor: .gray200,
-            text: "ESC",
+            tooltipShortcut: .keyboardShortcuts(.escape),
             tooltipPrompt: "Close Onit",
-            tooltipShortcut: .keyboardShortcuts(.escape)
+            text: "ESC"
```
timlenardo, line 33: undo shuffle here (reset the parameters to original order)

macos/Onit/UI/Content/ToolbarRight.swift
```diff
@@ -55,8 +55,8 @@ struct ToolbarRight: View {
             activeColor: .limeGreen,
             activeBackground: .clear,
             activeBorderColor: .clear,
-            tooltipPrompt: "Local Mode",
-            tooltipShortcut: .keyboardShortcuts(.toggleLocalMode)
+            tooltipShortcut: .keyboardShortcuts(.toggleLocalMode),
+            tooltipPrompt: "Local Mode"
```
timlenardo, line 59: undo shuffle here (reset the parameters to original order)

macos/Onit/UI/Prompt/Files/ContextItem.swift
```diff
@@ -31,7 +31,7 @@ struct ContextItem: View {
                     hoverBackground: isEditing ? (hasError ? .redDisabledHover : .gray400) : .gray600,
                     maxWidth: isEditing ? 155 : .infinity,
                     iconBundleURL: autoContext.appBundleUrl,
-                    tooltip: isEditing ? (hasError ? errorMessage : name) : "View auto-context file",
+                    tooltipPrompt: isEditing ? (hasError ? errorMessage : name) : "View auto-context file",
```
timlenardo, line 34: undo shuffle here (rename to 'tooltip')

macos/Onit/UI/Prompt/Files/ContextItem.swift
```diff
@@ -45,7 +45,7 @@ struct ContextItem: View {
                     maxWidth: isEditing ? 155 : .infinity,
                     iconView: ContextImage(context: item),
                     caption: item.fileType,
-                    tooltip: isEditing ? name : "View \(item.fileType ?? "") file",
+                    tooltipPrompt: isEditing ? name : "View \(item.fileType ?? "") file",
```
timlenardo, line 48: undo shuffle here (rename to 'tooltip')

macos/Onit/UI/Prompt/Files/FileRow.swift
```diff
@@ -86,7 +86,7 @@ extension FileRow {
                 hasHoverBorder: true,
                 shouldFadeIn: true,
                 iconBundleURL: iconBundleURL,
-                tooltip: "Add \(foregroundWindowName) Context"
+                tooltipPrompt: "Add \(foregroundWindowName) Context"
```
timlenardo, line 89: undo shuffle here (rename to 'tooltip')

macos/Onit/UI/Prompt/Files/WebContextItem.swift
```diff
@@ -49,7 +49,7 @@ struct WebContextItem: View {
             isLoading: websiteUndergoingScrape,
             iconView: websiteUndergoingScrape ? LoaderPulse() : favicon,
             caption: item.fileType,
-            tooltip: getCurrentWebsiteTitle(),
+            tooltipPrompt: getCurrentWebsiteTitle(),
```
timlenardo, line 52: undo shuffle here (rename to 'tooltip')

macos/Onit/UI/Components/ContextTag.swift
```diff
@@ -20,7 +20,7 @@ struct ContextTag: View {
     private let iconBundleURL: URL?
     private let iconView: (any View)?
     private let caption: String?
-    private let tooltip: String?
+    private let tooltipPrompt: String?
```
timlenardo, line 23: undo shuffle here (rename to 'tooltip')

timlenardo (changes requested):
This is looking good, but I have a few requests: 

1. Can you remove the 'ignoresMouseEvents' parameters everywhere? Instead, just set it to 'true' when creating the `tooltipWindow` in `setupTooltip()`.
2. Can you remove the renaming of "tooltip" to "tooltipPrompt" in ContextTag? 
3. In all the places where you moved lines of code around in a file without changing them, can you undo the changes? I'm marking them with 'undo shuffle here' comments below. 

Note: I acknowledge that requests 2 and 3 are atypically pedantic. In most cases, I won't request changes like this. However, as a team, we all need to work on simplifying our pull requests so that they're easier to review and ship. This PR (and #259!) provide good opportunities for practice, so I'm requesting the changes this time 😅

macos/Onit/UI/Components/TooltipView.swift
```diff
@@ -8,42 +8,106 @@
...
+        let textSubstrings: [String] = tooltip.prompt.components(separatedBy: newlineCharacters)
+        let textFirstLine: String = findValidSubstring(textSubstrings)
+        
+        let hasMultipleLines = textSubstrings.count > 1
+        
+        if hasMultipleLines {
+            VStack(alignment: .leading) {
+                Text(textFirstLine)
+                Text("...")
             }
```
timlenardo, line 48: As it's written, we only ever display the first line of text. That doesn't seem right. Shouldn't we respect the 'lineLimit' value set in the config? 

Both fixedTextView and trucatedTextView use this component, so all the logic around trimming the text and respecting the lineLimit isn't respected when we get here. "findValidSubstring' should be modified to remove any empty lines at the top, but otherwise not change the string. 

You can test this by:

1. Manually setting a long text value with multiple newlines ContextTag and 2. Setting a line limit in the ToolTip config. Here's my code (on line 136 of ContextTag): 

```
            TooltipHelpers.setTooltipOnHover(
                isHovering: isHovering,
                tooltipPrompt: "if !featureFlagsManager.usePinnedMode { \n Section { \n VStack(alignment: .leading, spacing: 4) { \n HStack { \n Text('Add Current Window on Launch') \n .font(.system(size: 13)) \n Spacer() \n Toggle('', isOn: $autoContextOnLaunchTethered) \n .toggleStyle(.switch) \n .controlSize(.small) \n SettingInfoButton( \n title: 'Add Current Window on Launch', \n description: \n When enabled, Onit automatically reads the tethered window and adds its text as context whenever the panel opens.', \n defaultValue: 'on', \n valueType: 'Bool' \n ) } \n Text('Adds context from the current window whenever Onit opens in Tethered mode.') \n .font(.system(size: 12)) \n .foregroundStyle(.gray200) \n } \n } \n }",
                tooltipConfig: TooltipConfig(maxWidth: 320, lineLimit: 4)
            )
```

When you run this, the lineLimit set in the tooltipConfig is ignored and we only display one line of text.

timlenardo (changes requested):

PR #315, opened 2025-07-09
Google search support

jrswanson:
- We're already supporting OpenAI and Anthropic native web search
- GoogleAI supports this, but requires us to use their native API format rather than OpenAI compatible Messages
- Migrated to the new format for both streaming and non-streaming endpoints
- Reused the `PerplexityEventParser` because it appeared to fix a similar bug with server event separators


timlenardo (approved):
Okay, let's get this merged, but with follow-up items for:

- Fix Citations
- Figure out how to show the 'search the web' animation in LLMStream
- Move logic gating web search for unsupported providers out of the UI element.

PR #317, opened 2025-07-10
Refresh button for remote models

timlenardo:
As part of the v2.8 release, I've added Grok 4 model. I was creating a demo video for the release notes and realized that there is no way to refresh the models list without restarting the app. I don't want to tell people to quit the app, so I added this simple refresh button, and was able to make a demo video using it: 

grok-4


jrswanson (approved):
LGTM

PR #339, opened 2025-07-16
Text Highlight: Update Highlighted Text Behavior

lk340:
This is part 1 in a series of mini-PRs that tackle various aspects of the overarching text highlight context experience. 
Please review this first before reviewing all other text highlight PRs.

Part 2: https://github.com/synth-inc/onit/pull/340

### Summary

* The `inputView` can now be permanently shown or hidden.
* `inputView` is permanently hidden when closing it through the newly-added cross ("x") button in the `inputView` title section.
* Highlighted text now appears as a context tag in `FileRow`.
* Clicking on the highlighted text context item reveals the `inputView`.
* `inputView` showing/hiding can also be toggled in the settings "General" tab, under the "Highlighted Text" section.

### Parts (Pin 1 and Pin 2 have been closed)

1. `YOU ARE HERE`
2. https://github.com/synth-inc/onit/pull/340
3. https://github.com/synth-inc/onit/pull/342 (Pin 1)
4. https://github.com/synth-inc/onit/pull/343 (Pin 2)
5. https://github.com/synth-inc/onit/pull/341 (Pin 3)

PR #340, opened 2025-07-16
Text Highlight: Auto-add To Context

lk340:
This is part 2 in a series of mini-PRs that tackle various aspects of the overarching text highlight context experience. 
Please review this after reviewing the [first PR](https://github.com/synth-inc/onit/pull/339).

### Summary

* This PR tackles the auto-adding nature of text highlights, namely togging the auto-adding of highlighted text into context on/off.
* Auto-adding behavior is set by the `autoAddHighlightedTextToContext` boolean property in the Defaults store (initialized to `true`).
* When `autoAddHighlightedTextToContext` is `true`, the app defaults to the original text highlight behavior.
* When `autoAddHighlightedTextToContext` is `false`, the highlighted text appears as the "ghost" button in `FileRow`, rather than automatically being added into context, and it takes precedence over the foreground window "ghost" button.
  * The "ghost" button tracks the highlighted text through the new `trackedPendingInput` state in `OnitPanelState`.
  * Clicking on the text highlight "ghost" button adds the highlighted text into context.
* The `autoAddHighlightedTextToContext` property can be toggled in the settings "General" tab, under the "Highlighted Text" section.
* Add a remove action to the highlighted text context tag, which sets `pendingInput` to `nil`.

### Parts

1. https://github.com/synth-inc/onit/pull/339
2. `YOU ARE HERE`
3. https://github.com/synth-inc/onit/pull/342 (Pin 1)
4. https://github.com/synth-inc/onit/pull/343 (Pin 2)
5. https://github.com/synth-inc/onit/pull/341 (Pin 3)

### DIFF
--- a/macos/Onit/UI/Prompt/Files/FileRow.swift
+++ b/macos/Onit/UI/Prompt/Files/FileRow.swift
@@ -14,6 +14,7 @@ struct FileRow: View {
     @ObservedObject private var debugManager = DebugManager.shared
     
     @Default(.autoContextFromCurrentWindow) var autoContextFromCurrentWindow
+    @Default(.autoAddHighlightedTextToContext) var autoAddHighlightedTextToContext
     
     @State private var ocrComparisonResult: OCRComparisonResult? = nil
     @State private var showOCRDetails = false
@@ -119,7 +120,7 @@ struct FileRow: View {
             FlowLayout(spacing: 6) {
                 PaperclipButton()
                 
-                addForegroundWindowToContextButton
+                addToContextButton
                 pendingWindowContextItems
                 highlightedTextContext
                 addedWindowContextItems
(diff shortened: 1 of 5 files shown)


macos/Onit/UI/Prompt/Files/FileRow.swift
```diff
@@ -150,28 +151,63 @@ struct FileRow: View {
...
+                let foregroundWindowName = WindowHelpers.getWindowName(window: foregroundWindow.element)
+                let iconBundleURL = WindowHelpers.getWindowAppBundleUrl(window: foregroundWindow.element)
+                
+                ghostContextTag(
+                    text: contextTagText,
+                    iconBundleURL: iconBundleURL,
+                    tooltip: "Add \(foregroundWindowName) Context"
+                ) {
+                    windowState?.addWindowToContext(window: foregroundWindow.element)
+                }
```
timlenardo, line 210: Let’s check with Elise when she’s online, but I would prefer to keep these as two separate ghost tags. While some designs have used a shared tag, I don’t think that’s the best user experience. Elise has mentioned that she often highlights text randomly, so hiding the main window ghost tag every time there is highlighted text seems like it will be irritating. Since the simplicity of adding the main window is a big value of our product, I think we should always keep that front and center!

timlenardo (approved):
This looks good, though we still need to switch the ordering of the Ghost Tags as discussed in our last meeting!

PR #341, opened 2025-07-17
Text Highlight (Pin - 3): FileRow Update

lk340:
### Summary

This PR *only* tackles the FileRow aspect of the pinned text highlight feature, but it includes all updates thus far. As a result, this branch *will* build and is ready for testing. I did this to separate concerns and make the updated code more reviewable.

### Parts

1. https://github.com/synth-inc/onit/pull/339
2. https://github.com/synth-inc/onit/pull/340
3.  (Pin 1)
4.  (Pin 2)
5. `YOU ARE HERE` (Pin 3)


macos/Onit/Data/Fetching/ChatEndpointMessagesBuilder.swift
```diff
@@ -46,13 +46,18 @@ struct ChatEndpointMessagesBuilder {
                     message += "\n\(webSearchContext.content)"
                 }
             }
+            
+            if !inputs[index].isEmpty {
+                let pendingInputs = inputs[index]
 
-           if let input = inputs[index], !input.selectedText.isEmpty { 
                 message += "\n\nUse the following selected text as context. When present, selected text should take priority over other context."
```
timlenardo, line 53: We should also update this to respect the ordering of the selected text. For example, if there's 4 highlighted text sections pinned and they send a prompt like "rephrase this", I think they're probably referring to the most recently added highlight. 

So this prompt should be modified to say like "the selected text should be given priority in the order they appear", and then we need to make sure that the most recent highlighted text is added first (which probably isn't the default, since we most likely use 'append').

macos/Onit/KeyboardShortcuts/KeyboardShortcutsManager.swift
```diff
@@ -106,8 +106,8 @@ struct KeyboardShortcutsManager {
                     if state.panel != nil {
                         if state.showContextMenuBrowserTabs {
                             state.showContextMenuBrowserTabs = false
-                        } else if state.pendingInput != nil {
-                            state.pendingInput = nil
+                        } else if state.hasHighlightedText {
+                            state.clearHighlightedTextStates()
```
timlenardo, line 110: Interesting. I wonder if this should only clear out the most recent, instead of all of them. It seems like it might be irritating if you hit Escape by accident and it clears your entire context window. 

Without really playing with it, my intuition is that we should: 
1. Only do this if the user is in the mode where highlights get auto-added. 
2. Only do it for the automatically added highlight, not any that they've pinned.

macos/Onit/UI/Components/ContextTag.swift
```diff
@@ -21,10 +21,12 @@ struct ContextTag: View {
     private let borderColor: Color?
     private let iconBundleURL: URL?
     private let iconView: (any View)?
+    private let iconViewCornerIcon: ImageResource?
```
timlenardo, line 24: I think I'd probably call this the 'leftIcon' and 'rightIcon' instead of 'iconView' and 'iconViewCornerIcon'. That's what this new field is, right? It's a new right-aligned icon?

macos/Onit/UI/Components/ContextTag.swift
```diff
@@ -41,10 +43,12 @@ struct ContextTag: View {
         borderColor: Color? = nil,
         iconBundleURL: URL? = nil,
         iconView: (any View)? = nil,
+        iconViewCornerIcon: ImageResource? = nil,
         caption: String? = nil,
         tooltip: String? = nil,
         errorDotColor: Color? = nil,
         action: (() -> Void)? = nil,
+        pinAction: (() -> Void)? = nil,
```
timlenardo, line 51: Calling this 'pinAction' makes it specific to pinning, which defeats the purpose of a general name like 'iconViewCornerIcon'. If we're only supporting a pin action, there's no need to pass in the CornerIcon; it can just be hardcoded. Alternatively, if we want to make this general, we should give both new fields corresponding generic names, like "rightIcon" and "rightIconAction." Either approach works for me since we have no immediate plans for anything other than pinning, but it should be one or the other.

macos/Onit/Data/Persistence/Prompt.swift
```diff
@@ -12,7 +12,7 @@ import SwiftData
 
     var instruction: String
     var timestamp: Date
-    var input: Input?
+    var inputs: [Input]
```
timlenardo, line 15: Were you able to get this code to run on your computer? It's insta-crashing for me. We can't change anything in any of the SwiftData model objects (Prompt, Chat, SystemPrompt, etc - search for the @model decorator) without writing a migration. In this case, transitioning from an optional Input to a non-optional input array results in a fatal error. Many of my old chats have a value of 'nil' for the 'input', which can't be converted into a non-optional array. 

To reproduce the issue, checkout an older version of the app and send a prompt with no Input. Then, check out this branch and run it. You'll get a fatal error. 

You can see some examples of migrations I've written in the past ('maybeUpdatePromptPriorInstructions' and 'maybeCleanupHangingPromptReferences') in SwiftDataContainer. It's a pain in the ass, but you have to do it if you want to change anything in Swift Data.

macos/Onit/UI/Components/ContextTag.swift
```diff
@@ -103,7 +113,21 @@ struct ContextTag: View {
...
+                        if let cornerIcon = iconViewCornerIcon {
+                            ZStack(alignment: .center) {
+                                Circle()
+                                    .fill(isHoveredBody ? hoverBackground : background)
+                                    .frame(width: 13, height: 13)
+                                
+                                Image(cornerIcon)
+                                    .addIconStyles(iconSize: 7.45)
+                            }
+                            .offset(x: 4, y: 4)
```
timlenardo, line 128: From reading the code, I can't tell the difference between this icon and the hoverActionButton added below. Do we need both?

timlenardo (changes requested):
Hey @lk340 - thanks for getting this up! Unfortunately, I am unable to run it due to a crash that occurs when I try to instantiate the data container, now that the SwiftData has changed. 

Fixing this is going to be a big task, and it feels like a niche product. How many people will really want to pin lots of highlighted text? It doesn't seem like a major user behavior, so I think we should set this aside for now and revisit it after we've addressed some more pressing matters. 

Regardless, I left some comments so you can understand the issues.

PR #347, opened 2025-07-18
Correctly resize window when panel has been shrunk

timlenardo:
This fixes another bug related to resizing other windows in Pinned mode: specifically, when the panel is reduced in size. 

Our resizing logic detects windows that are bordering or underneath the current panel. When the panel is shrunk, the windows are no longer bordering the OnitPanel, so they don't get resized. This creates an annoying problem where windows are left hanging. 

https://github.com/user-attachments/assets/dcd52203-1831-4323-8466-7020495ffd5a

The solution is to pass the original panel width to the resize function after resizing the panel. Then, use the larger value between the original and current panel widths to decide on window resizing. This step should be performed both in the bordering windows and again in all the windows. 

After this change, the issue is resolved: 

https://github.com/user-attachments/assets/b214a178-8cfc-442c-8297-2283d3e50f3b

### DIFF
--- a/macos/Onit/PanelStateManager/Pinned/PanelStatePinnedManager+Resize.swift
+++ b/macos/Onit/PanelStateManager/Pinned/PanelStatePinnedManager+Resize.swift
@@ -9,15 +9,15 @@ import AppKit
 
 extension PanelStatePinnedManager {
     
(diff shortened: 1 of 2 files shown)


Niduank (approved):
Let's merge it 👍

PR #348, opened 2025-07-18
Correctly remove pendingInput in Notes and other HighlightedTextWorke…

timlenardo:
…r apps

There was an edge case bug in the apps where we used HighlightedTextWorker instead of AXSelectedTextChanged notifications: when the user clicked away and deselected the text, the pendingInput would remain visible instead of disappearing.

https://github.com/user-attachments/assets/d9f9a26b-15f2-4427-8f7e-83ac05351cbd

This fixes it. In our selectionChangeHandler for the highlightedTextCoordinator, we need to call 'processSelectedText(nil)' when there is no element with selectedText. After this change, the pendingInput goes away as expected: 

https://github.com/user-attachments/assets/c05405de-a7ed-415c-86e1-20e9cf6db1b6


Niduank (approved):
Everything looks good to me 👍

PR #349, opened 2025-07-18
make default tooltip display match new helper

timlenardo:
I’ve noticed that some tooltips (such as on the ‘panel close’ button - aka TetheredButton.swift) appear immediately when you hover over them, which can be a bit irritating. Let's implement a standard delay across the app. We’re already using a delay in @lk340’s new tooltip helper, so let’s apply the same behavior to the .tooltip modifier. Unless a different delay is explicitly set, tooltips should appear with at least a 0.4-second delay.

@lk340, do you foresee any issues with this approach? Are there any elements where we’d want the tooltip to appear immediately instead?

### DIFF
--- a/macos/Onit/Helpers/TooltipHelpers.swift
+++ b/macos/Onit/Helpers/TooltipHelpers.swift
@@ -22,15 +22,10 @@ struct TooltipHelpers {
                         prompt: tooltipPrompt!,
                         shortcut: tooltipShortcut
                     ),
-                    tooltipConfig: tooltipConfig,
-                    delayStart: 0.4,
-                    delayEnd: 0
+                    tooltipConfig: tooltipConfig
                 )
             } else {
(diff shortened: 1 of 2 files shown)


timlenardo:
Hey @lk340 - thanks for the review! 

On your first point: `We can remove delayStart: 0.4 in TooltipHelpers.setTooltipOnHover()` - agreed! I'll remove this. 

On your second point:  `the tooltip on the TetheredButton fades out when hovering off of it` - The animation looks good, but it should be faster. We should also separate the animation from the delayEnd, as they are different concepts. I’ve added an ‘animateOut’ parameter, which defaults to true, and made the animation duration a static value with a default of 0.10. This should help maintain consistent animation timing across all tooltips. If we want to customize this in the future, we can make the duration configurable as a parameter.

Issue #316, opened 2025-07-10
Dock icon


freeshare:
hello, 

is it possible to have an option to "hide" the dock icon, and only keep the menu bar icon active?

Thank you

PR #351, opened 2025-07-21
Add setting that allows users to hide/show dock icon

timlenardo:
Requested in #316  - this PR allows users to hide/show the dock icon: 

https://github.com/user-attachments/assets/893f0c9d-d130-43fc-b0a4-7701b3e7058d

Under the hood, this toggles whether the app's 'ActivationPolicy' is in .accessory or .regular mode. 

I am concerned that switching it to .accessory will have some unforeseen consequences. @Niduank - you handled switching us from .accessory to .regular before. Did you have to change anything to make it work? 

 We should all use the app in accessory mode for a while to ensure everything is working correctly before releasing it.

### DIFF
--- a/macos/Onit/AppDelegate.swift
+++ b/macos/Onit/AppDelegate.swift
@@ -5,6 +5,7 @@
 //  Created by Kévin Naudin on 21/01/2025.
 //
 
+import Defaults
 import FirebaseCore
 import PostHog
 import SwiftUI
@@ -18,6 +19,10 @@ class AppDelegate: NSObject, NSApplicationDelegate {
             NSApp.appearance = darkAppearance
         }
         GIDSignIn.sharedInstance.restorePreviousSignIn()
+        
+        // Configure dock icon visibility based on user preference
+        configureDockIconVisibility()
+        
         // This is helpful for debugging the new user experience, but should never be committed!
         //        if let appDomain = Bundle.main.bundleIdentifier {
         //            UserDefaults.standard.removePersistentDomain(forName: appDomain)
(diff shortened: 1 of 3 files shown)


Niduank:
@timlenardo 
Setting the app to .accessory only affects its appearance in:
- The Dock,
- CMD+Tab,
- And the top-left menu bar.

Back then, I had to add the "LSUIElement" key in the Info.plist to force the app into accessory mode but `NSApp.setActivationPolicy` is actually enough!

The PR looks good, we can avoid code duplication and make a `static func toggleDockIconVisibility` in `App` or `AppDelegate`

macos/Onit/AppDelegate.swift
```diff
@@ -37,4 +42,18 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         
         return true
     }
+    
+    private func configureDockIconVisibility() {
```
Niduank, line 46: In order to avoid code duplication:
```swift
static func toggleDockIconVisibility()
``` 
and call in GeneralTab : `AppDelegate.toggleDockIconVisibility()`

Niduank (approved):

macos/Onit/AppDelegate.swift
timlenardo: Okay, I'll try this. I started with something similar but was having casting issues. The 'AppDelegate' available in views like GeneralTab is a generic instance, rather than our specific override. I think using a static function, as you suggest, should solve it though.

timlenardo: Making the function static worked well, good suggestion @Niduank 👍

timlenardo:
In addition to addresses the comments, I also decided to move the settings toggle from the "Auto-start" section (where it didn't make much sense), into the Display section. Ready to merge now!

PR #355, opened 2025-07-22
Handle streaming with Tools

Niduank:


timlenardo:
Awesome! @jrswanson - can you review this since you've also worked on tools? 

@Niduank - One issue I noticed off the top of my head:  PR #344 adds support for Google Tool calling. Also [Ollama supports tool calling too](https://ollama.com/blog/tool-support), so it'd be ideal to extend this to the local endpoint as well!

macos/Onit/Data/Tools/ToolProtocol.swift
```diff
@@ -0,0 +1,14 @@
...
+//
+//  Created by Kévin Naudin on 07/22/2025.
+//
+
+protocol ToolProtocol {
+    typealias ToolCallResultAlias = Result<ToolCallResult, ToolCallError>
+    
+    var activeTools: [Tool] { get }
+    func canExecute(partialArguments: String) -> Bool
+    func execute(toolName: String, arguments: String) async -> ToolCallResultAlias
```
timlenardo, line 13: Is the idea now that "execute" can be called multiple times as the response is streamed? For example, in the diff view, we will call "execute" repeatedly as the edited response is streamed. If so, we might want to separate this into an 'executePartial' and an 'executeFinal,' and each tool can have a 'canExecutePartial' flag. For some tools like Diff Tool, we'll want to execute partial responses. However, for other tools, such as CalendarTool's 'create_event', it doesn't make sense to execute partial responses.

macos/Onit/Data/Tools/ToolRouter.swift
```diff
@@ -8,36 +8,63 @@
...
+        guard let tool = tool else {
             return .failure(
-                ToolCallError(toolName: toolName, message: "Unrecognized app name: \(appName)"))
+                ToolCallError(toolName: toolName, message: "Unrecognized app name: \(appName)")
+            )
+        }
+        
+        if !isComplete && !tool.canExecute(partialArguments: toolArguments) {
+            return nil
+        } else {
```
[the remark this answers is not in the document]
Niduank: `nil` is used when the tool can't be executed with partial arguments. It's neither an error nor a success.

macos/Onit/Data/Tools/ToolProtocol.swift
Niduank: @timlenardo 
That's exactly what I did, except I didn't see the need to create two separate execute functions.
By conforming to the `ToolProtocol`, each tool must implement a method `func canExecute(partialArguments: String) -> Bool`.
If this function returns false, the tool is not executed (it's neither a success nor an error), which is the case for the CalendarTool.
In the case of the DiffTool, the arguments can be partial, so we simply wait until at least the required argument "improved_content" is present.

jrswanson:
> > Awesome! @jrswanson - can you review this since you've also worked on tools?
> > @Niduank - One issue I noticed off the top of my head: PR #344 adds support for Google Tool calling. Also [Ollama supports tool calling too](https://ollama.com/blog/tool-support), so it'd be ideal to extend this to the local endpoint as well!
> 
> I had two questions for @jrswanson, but @timlenardo beat me to it, so I only have one left: How does the Onit server handle tool streaming?

Currently it just passes along the partial json for OpenAI and Anthropic. Google's API doesn't do partial json so this is passed back whole.

jrswanson (approved):

PR #360, opened 2025-07-25
Introduce keystroke notification manager

timlenardo:
This is the first step of many towards merging some of my typeahead test collection code. This PR introduces the KeystrokeNotificationManager, the KeystrokeNotificationDelegate, and the KeycodeTranslator. The first two allow us to listen for keystroke events. It's relatively simple for now; we are currently only listening for key-down events and then passing them along to any delegate. 

KeycodeTranslator is a helper method that translates keycodes into readable characters and also determines if the keys produce an output or not.

Note: these classes are not used anywhere in the codebase in this PR. I've tested them in my more extensive typeahead branch, and all seems to be working as expected!

### DIFF
--- a/macos/Onit/Mouse/MouseNotificationDelegate.swift
+++ b/macos/Onit/Mouse/MouseNotificationDelegate.swift
@@ -1,0 +1,19 @@
+//
+//  MouseNotificationDelegate.swift
+//  Onit
+//
+//  Created by Timothy Lenardo on 7/25/25.
+//
+
+import Foundation
+import AppKit
+
+@MainActor protocol MouseNotificationDelegate: AnyObject {
+    func mouseNotificationManager(_ manager: MouseNotificationManager, didReceiveSingleClick event: NSEvent)
+    func mouseNotificationManager(_ manager: MouseNotificationManager, didReceiveDoubleClick event: NSEvent)
+    func mouseNotificationManager(_ manager: MouseNotificationManager, didReceiveTripleClick event: NSEvent)
+    func mouseNotificationManager(_ manager: MouseNotificationManager, didStartDrag event: NSEvent)
+    func mouseNotificationManager(_ manager: MouseNotificationManager, didUpdateDrag event: NSEvent)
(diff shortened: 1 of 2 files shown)


macos/Onit/Keystrokes/KeycodeTranslator.swift
```diff
@@ -0,0 +1,253 @@
...
+
+                return textModifyingControlKeys.contains(keyCode)
+            }
+
+            // Shift and Option with output keys still produce output (shifted chars, accented chars)
+            return true
+        }
+
+        // For unknown keys, assume they don't modify text field contents
+        return false
```
Niduank, line 209: We should extract all the `___Keys: Set<Int64>` in the class so the function will be more readable.

macos/Onit/Keystrokes/KeystrokeNotificationManager.swift
```diff
@@ -0,0 +1,191 @@
...
+
+    private func notifyDelegates(_ notification: (KeystrokeNotificationDelegate) -> Void) {
+        for case let delegate as KeystrokeNotificationDelegate in delegates.allObjects {
+            notification(delegate)
+        }
+    }
+
+    // MARK: - Callbacks
+
+    private var keystrokeCallbacks: [(KeystrokeEvent) -> Void] = []
```
Niduank, line 62: You already use a delegate pattern to notify about keystroke event. 
So the closure pattern isn't necessary.

Niduank (changes requested):
First of all, thanks for splitting your architecture into multiple PRs. Really appreciate the idea !

There are a few improvements we can make to the code.

What I was trying to explain yesterday during the meeting with my Frenglish:
I'm using an AZERTY keyboard, while you're on QWERTY, so I'm not sure your code will work for me.

I had issues in the past with keyboard layout differences when working with CGKeyCode to simulate a CMD+A in the ClipboardParser. I had to build a custom transformation for that in String+KeyCode.swift.

You can easily test this by switching your keyboard layout.

PR #362, opened 2025-07-25
Feature/mouse notification manager

timlenardo:
Building off PR #360 , this PR introduces a Mouse notification manager. Currently, it can only send a few events (single, double, & triple clicks plus drags) and we aren't using it anywhere, but we will want this soon in a few cases:

- Our nextgen highlighted text detection will need to listen for double/triple clicks and drags, because the kAXSelectedTextNotifications aren't reliable. 
- Cursor position detection (useful for the Typeahead UI & QuickEdit hint) will use single clicks to detect when the cursor has moved. 
- And potentially more in the future!

### DIFF
--- a/macos/Onit/Keystrokes/KeycodeTranslator.swift
+++ b/macos/Onit/Keystrokes/KeycodeTranslator.swift
@@ -2,252 +2,214 @@
 //  KeycodeTranslator.swift
 //  Onit
 //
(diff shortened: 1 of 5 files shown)


Niduank (approved):
@timlenardo Code looks good !
One thing to have in mind: `globalEventMonitor` works with accessibility's permission.
If permission is denied, `addGlobalMonitorForEvents` will silently fail.

PR #363, opened 2025-07-26
Setup the ability to listen for AX notifications from our own process…

timlenardo:
… and ignored processes

This is the first of many PRs related to the upcoming typeahead system. For typeahead, we want to be able to listen for some notifications on both our own process and the ignored process. This PR modifies the AccessibilityNotificationsDelegate so that it can ask for notifications from either by setting `wantsNotificationsFromIgnoredProcesses` or `wantsNotificationsFromOnit` to true. 

Note:
- The .nonactivatingPanel styleMask on OnitRegularPanel prevents standard “app activated” and “deactivated” notifications. This makes it difficult to start and stop the accessibilityNotificationsManager based on Onit’s active state. Removing the .nonactivatingPanel caused significant UI issues, so as a workaround, we now continuously listen for a small subset of AXNotifications from our own process.

### DIFF
--- a/macos/Onit/Accessibility/Observers/AccessibilityObserversManager.swift
+++ b/macos/Onit/Accessibility/Observers/AccessibilityObserversManager.swift
@@ -21,6 +21,15 @@ class AccessibilityObserversManager {
     // MARK: - Singleton
     
     static let shared = AccessibilityObserversManager()
+
+    // Static method to get the current ignored app names (for consistency across managers)
+    static var currentIgnoredAppNames: [String] {
+    #if DEBUG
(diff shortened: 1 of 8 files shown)


macos/Onit/Accessibility/Observers/AccessibilityObserversManager.swift
```diff
@@ -345,6 +402,15 @@ class AccessibilityObserversManager {
         return .eligible
     }
     
+    // Static method to get the current ignored app names (for consistency across managers)
```
Niduank, line 405: Let's move this variable at the beginning of the file.

PR #364, opened 2025-07-28
Logged Out Mode: Auth Manager

lk340:
### Summary

This PR addresses the following:

1. Create `AuthManager` to consolidate auth logic that is scattered across various places within the app into a single source of truth.
2. Further consolidate provider-based logic inside of `ModelProvidersManager` to allow for better UI reactivity while improving the single source of truth paradigm for state-based model provider variables and methods.
3. Provide a small fix for the UI not properly accounting for custom providers when enabling and disabling models.
4. Provide a small fix for `OnboardingSkipAccesssibility` forcing the auth flow to show, even when logged in.


macos/Onit/UI/Auth/AuthFlow.swift
```diff
@@ -24,6 +26,8 @@ struct AuthFlow: View {
     @Default(.useGoogleAI) var useGoogleAI
     @Default(.useDeepSeek) var useDeepSeek
     @Default(.usePerplexity) var usePerplexity
+    
+    @State private var modelProvidersManager = ModelProvidersManager.shared
```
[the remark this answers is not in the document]
lk340: Same as above

macos/Onit/AppState.swift
```diff
@@ -79,12 +62,32 @@ class AppState: NSObject, SPUUpdaterDelegate {
     var showFreeLimitAlert: Bool = false
     var showProLimitAlert: Bool = false
     var subscriptionPlanError: String = ""
+    
+    private var authCancellable: AnyCancellable? = nil
```
jrswanson, line 66: Does this binding get cancelled somewhere?

macos/Onit/ModelProvidersManager/ModelProvidersManager.swift
```diff
@@ -95,4 +159,55 @@ class ModelProvidersManager {
...
+            return !token.isEmpty && isGoogleAITokenValidated
+        case .deepSeek:
+            guard let token = deepSeekToken else { return false }
+            return !token.isEmpty && isDeepSeekTokenValidated
+        case .perplexity:
+            guard let token = perplexityToken else { return false }
+            return !token.isEmpty && isPerplexityTokenValidated
+        case .custom:
+            return false /// When checking for valid tokens on custom providers, use `hasValidCustomToken()` below.
+        }
```
jrswanson, line 203: Would returning `nil` here make more sense because the truth value is undetermined?

lk340: Yup, in my previous implementation of `ModelProvidersManager`'s methods, I'd have argued for this! While rewriting the code logic, however, something I'd noticed is that this could result in having do a double-take on checking for valid remote provider tokens when referencing `ModelProvidersManager`... :

1. Checking for a "standard" remote provider's (e.g. OpenAI, Anthropic, etc.) token.
2. And then checking for a "custom" remote provider's token, in the case where the above returns `nil`.

...which doesn't strike me as being readable and simple. So, rather than returning `nil`, I've rewritten the class methods to handle the check logic internally within the class itself, so outside entities who reference `ModelProvidersManager` only need to call a single method.

jrswanson (approved):

PR #368, opened 2025-07-29
Add HighlightedTextManager and pull highlighted text logic into it

timlenardo:
This is part 3 of my refactoring series. 

In this PR, I pull the highlighted text logic out of the AccessibilityNotificationManager and into a new HighlightedTextManager. This will be necessary as we add other ways to detect highlighted text in the future. For instance, when detecting highlighted text with vision models in non-accessibility apps like Google Docs, we'll need that logic to be outside the AccessibilityNotificationManager, because it will use notifications like mouse drags, or shift+arrow key presses. 

This also introduces the pattern that I have advocated for in a few of our engineering meetings: there should be "state managers" that are in charge of calculating specific states that we are interested in, like which text is highlighted, or which window is foregrounded. These need to be separate from the notification managers, like AccessibilityNotificationManager, the MouseNotificationManager, and the KeystrokeNotificationManager, because we'll often need to use notifications from different managers to determine the correct answer. These states should also be outside of the OnitPanelState, since we'll want them in different features, like QuickEdit or Typeahead. This new pattern accomplishes both.

Note: I've left the code 'as is' to make the review process easier. However, the HighligtedTextManager code now contains some code related to the CaratPositionManager, which should be stripped out. I will do this in a follow-up PR.

### DIFF
--- a/macos/Onit/Accessibility/Notifications/AccessibilityNotificationsManager.swift
+++ b/macos/Onit/Accessibility/Notifications/AccessibilityNotificationsManager.swift
@@ -163,7 +163,6 @@ class AccessibilityNotificationsManager: ObservableObject {
             self.handleSelectionChange(for: element)
         case kAXValueChangedNotification:
             self.handleValueChanged(for: element)
-            // self.handleCaretPositionChange(for: element)
         case kAXFocusedUIElementChangedNotification:
             self.handleFocusedUIElementChanged(for: element)
         case kAXWindowMovedNotification:
--- a/macos/Onit/StateManagers/HighlightedText/HighlightedTextManager.swift
+++ b/macos/Onit/StateManagers/HighlightedText/HighlightedTextManager.swift
@@ -22,6 +22,7 @@ class HighlightedTextManager: ObservableObject {
     private var lastHighlightingProcessedAt: Date?
     private var selectionDebounceWorkItem: DispatchWorkItem?
     private var currentSource: String?
+    private var lastCaretPositionChangeTimestamp: Date?
     
     // Published property for selected text that QuickEditManager can observe
     @Published var selectedText: String?
@@ -130,10 +131,4 @@ class HighlightedTextManager: ObservableObject {
         selectionDebounceWorkItem?.cancel()
         selectedText = nil
     }
-    
-    // MARK: - Private Properties
-    
-    private var lastCaretPositionChangeTimestamp: Date?
-    
-
 }
--- a/macos/Onit/Accessibility/Notifications/AccessibilityNotificationsManager+Config.swift
+++ b/macos/Onit/Accessibility/Notifications/AccessibilityNotificationsManager+Config.swift
@@ -12,6 +12,6 @@ extension AccessibilityNotificationsManager {
     struct Config {
 
         static let debounceInterval: TimeInterval = 0.3  // 300ms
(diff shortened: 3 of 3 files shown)


macos/Onit/StateManagers/HighlightedText/HighlightedTextManager.swift
```diff
@@ -0,0 +1,139 @@
...
+    
+    func reset() {
+        lastHighlightingProcessedAt = nil
+        selectionDebounceWorkItem?.cancel()
+        selectedText = nil
+    }
+    
+    // MARK: - Private Properties
+    
+    private var lastCaretPositionChangeTimestamp: Date?
```
lk340, line 136: This should be moved to the top of the file with the other private properties for better readability

macos/Onit/StateManagers/HighlightedText/HighlightedTextConfig.swift
```diff
@@ -0,0 +1,12 @@
...
+//  HighlightedTextConfig.swift
+//  Onit
+//
+//  Created by Kévin Naudin on 28/04/2025.
+//
+
+import Foundation
+
+struct HighlightedTextConfig {
+    static let textSelectionDebounceInterval: TimeInterval = 0.15  // 150ms
```
lk340, line 11: We should remove the vestigial `textSelectionDebounceInterval` from `AccessibilityNotificationsManager+Config.swift`.

macos/Onit/Accessibility/Notifications/AccessibilityNotificationsManager.swift
```diff
@@ -178,7 +165,7 @@ class AccessibilityNotificationsManager: ObservableObject {
             self.handleValueChanged(for: element)
             // self.handleCaretPositionChange(for: element)
```
lk340, line 166: Maybe this comment is needed here as a future reminder, but I'm just pointing out that it's here and should maybe be removed if it's safe!

lk340 (approved):

lk340:
Forgot to add a quick comment about this, but we should also make the methods inside of `HighlightedTextManager` private to signify that these operations are isolated to this class.

macos/Onit/Accessibility/Notifications/AccessibilityNotificationsManager.swift
```diff
@@ -136,10 +124,9 @@ class AccessibilityNotificationsManager: ObservableObject {
                 
                 Task { @MainActor in
                     if let element = unsafeElement {
-                        self.processSelectionChange(for: element)
+                        self.handleSelectionChange(for: element)
                     } else {
-                        // Handle text deselection - clear pendingInput
-                        self.processSelectedText(nil)
+                        HighlightedTextManager.shared.processSelectedText(nil)
```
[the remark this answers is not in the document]
timlenardo: @lk340 - the issue is in `notifyDelegates` in AccessibilityNotificationsManager, where we check the processID before sending the notification. If we have a nil AXUIElement, we won't know which process it's from, so we will always send the notification. That would be okay for now, since only Notes and iTerm2 use the HighlightedTextPoller. But it would break if want to add any app using the HighlightedTextPoller to the ignoredApps list. In this case, we'd send the notification to all delegates, even if the app _should_ be ignored.

PR #369, opened 2025-07-29
Move pendingInput code to OnitPanelState

timlenardo:
Part 4 of my refactoring series..

This PR pulls the code for updating the PendingInput to the OnitPanelState. We introduce a HighlightedTextDelegate that allows the HighlightedTextManager to send notifications when the highlighted text changes. The OnitPanelState can then subscribe to these notifications and own its logic for updating the PendingInput, disentangling the two. 

One note: this might create some weird behavior in Tethered mode, since it will update ALL of the OnitPanelStates, instead of only the one in PanelStateCoordinator's state object. However, I think this is the better solution? If you highlight text, I would expect it to be available in all of the OnitPanels. 

Next up: removing the Caret management code from the HighlightedTextManager.


Niduank (approved):

lk340 (approved):
LGTM! Loving this refactor series so far! 💪

